PR #20, opened 2021-09-22
Update release-notes-21-08.adoc

Pubs-MV:
updated the versioned API to clarify what it means to use v1 vs the versioned APIs.

## Types of changes

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes if appropriate.
- [ ] All new and existing tests passed.

### DIFF
--- a/rn/release-information/release-notes-21-08.adoc
+++ b/rn/release-information/release-notes-21-08.adoc
@@ -71,12 +71,14 @@ Starting in Kepler (the release after Joule), we'll offer full n-2 backwards com
 ==== Versioned API
 
 // #26281
-* With each release there will now be a versioned API and the API version is specified in the URI.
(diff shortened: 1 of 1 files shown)


Aqsaxvv092 (changes requested):
"The v1 version always points to the latest API and you can continue to use the v1 APIs. " This is still very confusing in the way it is worded. /v1 api doesn't automatically point to "latest" version available. It simply points the version that the Console is on. Please see if we can use the verbiage from my recommendation in doc.

Pubs-MV:
> "The v1 version always points to the latest API and you can continue to use the v1 APIs. " This is still very confusing in the way it is worded. /v1 api doesn't automatically point to "latest" version available. It simply points the version that the Console is on. Please see if we can use the verbiage from my recommendation in doc.

Please take a look and edit it if you think we need to revise it more.

iansk (approved):

PR #27, opened 2021-10-04
Made clear how serverless waas is licensed

avish9:
Serverless WAAS is counted and licensed as part of the Serverless licensing.


iansk:
Thanks @avish9. Merged this now

PR #45, opened 2021-10-14
Update release-notes-21-08-update1.adoc

Pubs-MV:
added known issue 33004


Pubs-MV:
approved by Mor

MorShabiPAN:
Hi @Pubs-MV I think we should remove the "or later" from the description

PR #48, opened 2021-10-18
twistcli_scan_IAC is now deprecated.

kdpanw:
This feature is deprecated / removed with the recent update see - https://docs.paloaltonetworks.com/prisma/prisma-cloud/21-08/prisma-cloud-compute-edition-release-notes/release-information/release-notes-21-08.html

Because of this, have had customers open cases confused on the proper way to do this.
It appears we can still do this operation by using the API itself.
Suggesting that perhaps instead of have the page above have the info it holds, maybe just have a link to where to go etc.
Not entirely sure if this suggested change is the best method- but I still want to get the ball rolling on this. Thank you!!


iansk (approved):

iansk:
Thanks @kdpanw, good fix. Merging this now

kdpanw:
@iansk  Thank you very much!

PR #53, opened 2021-10-20
Added missing perms from serverless auto-defend

avish9:


iansk:
Thanks @avish9. Merging this now

PR #57, opened 2021-10-21
Added the reason for requiring Admin privileges

grevach:


grevach:
@iansk 
I'm pretty sure I did not phrase the sentence properly. I had a hard time finding a way to do this without making it too long.
It would be great if you could find better wording.

Thanks!

iansk:
Thanks @grevach. Looks good. I made some small adjustments to the wording. Merging this now.

PR #66, opened 2021-10-27
Update instructions to reflect that 'Downloads' is now 'Utilities'

gabe-sky:
We renamed the Downloads tab to Utilities.  This is an update to replace it in the docs, in several places.


Pubs-MV:
Thank you @gabe-sky

PR #69, opened 2021-10-27
Create SaaS Console Logs.adoc as a new Topic

kdpanw:
There is currently no documentation for this endpoint but it is in fact required in some cases. Please review and consider this PR as we currently have a customer who we had to describe this to, but it is not documented. Thank you

CC: Cliff Shih.

### DIFF
--- a/admin_guide/audit/SaaS Console Logs.adoc
+++ b/admin_guide/audit/SaaS Console Logs.adoc
@@ -1,0 +1,14 @@
+SaaS Console Log
+
+The following API call can be used to export console logs in the instance that the "syslog" feature is disabled for SaaS Compute.
+
+curl 'https://us-west1.cloud.twistlock.com/<Path to Console>/api/v1/logs/console?lines=1000' -X 'GET' \
+  -H 'accept: */*' \
+  -u  'access_key:access_token'
+
+
+  Where:
+
+  <Path to Console> = Path to console, found at 
+  api/v1/logs = is your endpoint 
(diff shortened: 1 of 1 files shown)


Pubs-MV:
Thank you for submitting a PR @kdpanw!  
You use api/v1/logs what should that point to in  the versioned endpoint list? I could not find it, in the API reference on https://prisma.pan.dev/api/cloud/cwpp

If it is supported, would users look for this in the API reference or better in the API workflow section on https://prisma.pan.dev/docs/cloud/cwpp/api-workflows/?

kdpanw:
I will update this part with the information as soon as I collect it!

kdpanw:
@Pubs-MV 

I believe that is the issue. That endpoint exists but it is not documented anywhere. I think it would look better in the API reference, with its own section perhaps. I suppose that is up to debate (I'm not entirely sure what would look better)

iansk:
Thanks, @kdpanw. Very much appreciate your submission here. For now, our policy is that we only document versioned (i.e. supported) API endpoints. We strongly encourage customers to use versioned endpoints only because of the long term support for them, and the backwards compatibility guarantees. For the v1 endpoints, we don't guarantee any sort of stability. v1 endpoints can change from release to release, and that can impact a customer's operations. For that reason, we don't recommend customers build on v1 endpoints, and, don't document them.

That doesn't mean v1 endpoints are inaccessible. You can still invoke them. And our OpenAPI file contains the complete spec for them. We just don't offer docs for them.

PR #70, opened 2021-10-27
update install_defender_gcp_marketplace.adoc

kdpanw:
The image is outdated for this, as well as the search query term. To find the correct application you must search Prisma Cloud - Kubernetes Security Defender.
None of the other options show the "configure" button as the documentation suggests.

I will also make a PR for the new graphic to use here shortly.


iansk:
Thanks @kdpanw. Good update. This is now merged

PR #75, opened 2021-10-31
Updates on cluster context following customer questions

grevach:
- Which field indicates the cluster name in the kubeconfig (server)
- Implication of changing the cluster name


iansk (approved):

iansk:
Thanks @grevach. This is merged. I will also publish it today.

PR #76, opened 2021-11-01
updated to remove confusion on supported tools

jjchavanne:
Reworded to be more generic and leave secrets integrations to the more detailed "Integrate with a secrets store" page.  The main page made it sound like we only supported 2 out of the current 5 options.


iansk:
@jjchavanne Very much appreciate the fix.

We recently did a big refactor of the repo, so this merge across forks was a pain. I replicated your changes with this commit:

https://github.com/PaloAltoNetworks/prisma-cloud-docs/commit/a0c7226d3cb3f1551a2161e56d47774c61e1f202

Closing this unmerged (but committed the changes separately, see above)

PR #94, opened 2021-11-15
Update releases.adoc

Pubs-MV:
fixed a typo: Iverson was 21.08 and not 20.08


cmalvia:
Please correct it..  For example, 20.08.514 is an Iverson release, which first shipped in August, 2021.

PR #100, opened 2021-11-17
Added the template book

arun728:
Added the template book - works fine now.


iansk:
Thanks @arun728. Merging this now

PR #106, opened 2021-11-18
Clarify WAAS "server" benchmark is WAAS-protected nginx

gabe-sky:
This commit updates the WAAS benchmark text to clarify that the "servers" compared are both nginx, just one of them has WAAS enabled.  It also spells out that negative "overhead" values indicate a speed-up, rather than assuming the reader knows that.


iansk:
@e-shuster Just a heads-up that I'm merging this change

iansk:
Thanks @gabe-sky

PR #110, opened 2021-11-22
Admission configuration

MorShabiPAN:
The current configuration flow is in current and case many configuration issues and bugs. The script provided by the UI is actually deprecated in v 1.22+ and not in v 1.16.

Therefore, I updated the note about the supported versions and updated the instruction to copy the configuration yaml file from the UI.

This fix is important and urgent since many customers raised a support case because of that

cc @MayaShani @iansk


MorShabiPAN:
Related to https://spring.paloaltonetworks.com/twistlock/twistlock/issues/23733

iansk:
Thanks @MorShabiPAN. Looks good, merging this now. I'll publish it a little later today.

PR #114, opened 2021-11-30
update post.md

kdpanw:
See Jira for more info : https://redlock.atlassian.net/browse/PCSUP-6947 . The collections POST endpoint does not instantiate correctly VIA api call . You must instantiate all values applicable as a * wildcard, in order for them to be added. They do not add by default.

### DIFF
--- a/api/descriptions/collections/post.md
+++ b/api/descriptions/collections/post.md
@@ -23,5 +23,6 @@ $ curl 'https://<CONSOLE>/api/v<VERSION>/collections' \
 }'
 ```
 Any resource left unspecified, such as `hosts`, `functions`, or `clusters`, is assigned a wildcard by default.
+Resources that are not explicitly specified need to have a `*` wildcard applied to them in order to instantiate.
 
-**Note:** No response will be returned upon successful execution.
+
--- a/api/descriptions/collections/post.md
+++ b/api/descriptions/collections/post.md
@@ -1,4 +1,4 @@
-Creates a new collection.
+Creates a new collection. You need to either specify a value or initialize the resouces in the endpoint with a "*" wildcard to be able to use the collection.
 
 To invoke this endpoint in the Console UI:
 
@@ -8,10 +8,10 @@ To invoke this endpoint in the Console UI:
 
 ### cURL Request
 
-Refer to the following example cURL command that creates a new collection named `my-collection` and captures all container images named `ubuntu:18.04`:
+Refer to the following example cURL command that creates a new collection named `my-collection`, specifies a HEX color value of #AD3C21, and captures all container images named `ubuntu:18.04`:
 
 ```bash
-$ curl 'https://<CONSOLE>/api/v<VERSION>/collections' \
+$ curl 'https://<CONSOLE>:8083/api/v<VERSION>/collections' \
   -k \
   -X POST \
   -u <USER> \
@@ -19,10 +19,19 @@ $ curl 'https://<CONSOLE>/api/v<VERSION>/collections' \
   -d \
 '{
    "name":"my-collection",
-   "images":["ubuntu:18.04"]
+   "images":["ubuntu:18.04"],
+   "hosts":["*"],
+   "labels":["*"],
+   "containers":["*"],
+   "functions":["*"],
+   "namespaces":["*"],
+   "appIDs":["*"],
+   "accountIDs":["*"],
+   "codeRepos":["*"],
+   "clusters":["*"], 
+   "color":"#AD3C21"
 }'
 ```
-Any resource left unspecified, such as `hosts`, `functions`, or `clusters`, is assigned a wildcard by default.
-Resources that are not explicitly specified need to have a `*` wildcard applied to them in order to instantiate.
+**Note:** Resources that are not explicitly specified need to have a `*` wildcard applied to them in order to instantiate as mentioned in the example.
 
 
--- a/api/descriptions/collections/post.md
+++ b/api/descriptions/collections/post.md
@@ -1,4 +1,10 @@
-Creates a new collection. You need to either specify a value or initialize the resouces in the endpoint with a "*" wildcard to be able to use the collection.
+Creates a new collection. You must specify a value for mandatory resources such as 'name', 'images', and 'labels' in the collection.
+
+For 'name': Valid characters are 'A-Z', 'a-z', '0-9', '_', '-', and ':'.
+For 'image': If you don't have a specific value, you can initialize with a wildcard '*'.
+For 'label': If you don't have a specific value, you can initialize with a wildcard '*'.
+
+Values for the mandatory resources are required for initialization. If you don't initialize mandatory resources and try to use the collection, you'll get an empty resource error.
 
 To invoke this endpoint in the Console UI:
 
@@ -32,6 +38,6 @@ $ curl 'https://<CONSOLE>:8083/api/v<VERSION>/collections' \
    "color":"#AD3C21"
 }'
 ```
-**Note:** Resources that are not explicitly specified need to have a `*` wildcard applied to them in order to instantiate as mentioned in the example.
+**Note:** No response is returned upon successful execution. You must verify the collection in the Console UI.
 
 


iansk:
@ssugandh Please validate the current content, and fix if necessary.

api/descriptions/collections/post.md
```diff
@@ -23,5 +23,6 @@ $ curl 'https://<CONSOLE>/api/v<VERSION>/collections' \
 }'
 ```
 Any resource left unspecified, such as `hosts`, `functions`, or `clusters`, is assigned a wildcard by default.
+Resources that are not explicitly specified need to have a `*` wildcard applied to them in order to instantiate.
```
ssugandh: The statement at line number 26 suggests some user action for adding a wildcard '*'. But, it's automatically gets added for instantiation as mentioned in line number 25. We can rephrase as "To instantiate a resource object, a wildcard '*' is assigned by default if the resource is left unspecified."

api/descriptions/collections/post.md
```diff
@@ -23,5 +23,6 @@ $ curl 'https://<CONSOLE>/api/v<VERSION>/collections' \
...
+
 
-**Note:** No response will be returned upon successful execution.
```
ssugandh, line 27: This is still valid. Not sure, why this got removed.

api/descriptions/collections/post.md
kdpanw: Hi, 
   If you create a collection via the API and do not specify a wildcard, it will not be added. I can demo this is necessary. That is why I suggest we remove the line saying it is added by default, as it is not.

api/descriptions/collections/post.md
kdpanw: There are situations where we receive no response due to perhaps a 404/400 error but the cURL has no response shown, and this indicates that it is successful. However, it is a failure. Running cURL -vvv on your API calls can show this in some aspects.

kdpanw: If you are certain that this functionality is behaving as expected, perhaps it is some bug with cURL or some other variable, and I can research that and see what I find and create a different PR/Report for that aspect.

api/descriptions/collections/post.md
ssugandh: Okay, I maybe wrong but during my testing when I created a collection via API, the wildcards got added as mentioned.

api/descriptions/collections/post.md
ssugandh: True, but we have several other instances similar to this. Shall we remove such statements from other API descriptions too?

kdpanw: I personally believe we should. I do not think there is any added value from having this statement, as it can confuse customers. It also is a blanket statement that doesn't necessarily cover every single use-case. Also, product team (I assume) is not keeping up with that as some sort of standard. What I do know, is that I have seen this occur where we have no response, but it's not completing the API call. It's getting a 404. I think that we get more benefit from removing this statement, than we do from having it in the docs. Generally, customers using the API are well-off enough to navigate around themselves and will know whether or not it works.

api/descriptions/collections/post.md
kdpanw: I am not sure what operation you performed or what test you ran to do this, it's a specific use-case. This may be easier to sync on with a quick huddle over slack if you'd like. Here, @ssugandh , take a quick look at my repo on this JIRA: https://redlock.atlassian.net/browse/PCSUP-6947 . I have some images in there that should shed some light on this.

ssugandh: I see that. The issue seems to be true. I verified as suggested in the Jira.

When I create a collection via API endpoint, the wildcards are added or atleast it seems to be added on the UI. I verified with the UI. But, when I use that same collection to create a rule, it throws that error like it's mentioned in the Jira. Unless, I pass a wildcard explicitly through either POST or PUT, it's treated as empty.

iansk: Excellent work, @ssugandh 

@kdpanw Does this cover the full extent of the issue?

iansk: This line has been updated. Resolving.

api/descriptions/collections/post.md
```diff
@@ -1,4 +1,4 @@
-Creates a new collection.
+Creates a new collection. You need to either specify a value or initialize the resouces in the endpoint with a "*" wildcard to be able to use the collection.
```
iansk: @ssugandh Couple of things:

1. Please use semantic lines breaks (i.e. one sentence per line). See the guidance here: https://sembr.org/

2. Please update this sentence to: 

   ```
   You must specify a value for all resources in the collection, even if it's just a wildcard `*`.
   ```

3. Add a new paragraph that describes why it's important to specify a value for all resources, and what happens if you don't do that.

   ```
   If you create a collection and specify values for only some resources, then ... <WHAT'S THE OUTCOME, WHY THIS IS IMPORTANT, HOW YOU CAN TELL YOU DID SOMETHING WRONG, HOW TO VALIDATE, ETC>
   ```

api/descriptions/collections/post.md
```diff
@@ -8,20 +8,30 @@ To invoke this endpoint in the Console UI:
 
 ### cURL Request
 
-Refer to the following example cURL command that creates a new collection named `my-collection` and captures all container images named `ubuntu:18.04`:
+Refer to the following example cURL command that creates a new collection named `my-collection`, specifies a HEX color value of #AD3C21, and captures all container images named `ubuntu:18.04`:
```
iansk, line 17: @ssugandh What happens if you don't specify a color? Does the system randomly assign the collection a unique color? Please test it in your lab setup. If it does, then remove this param from the example curl command.

I want our example curl commands to show only the **minimum params** do required to do something **useful in the real world**. (Sometimes our API lets you specify even fewer params then we show in our curl commands, but the created object doesn't do anything)

api/descriptions/collections/post.md
iansk: @ssugandh What is the response code when a collection is successfully created? 200?

api/descriptions/collections/post.md
```diff
@@ -8,20 +8,30 @@ To invoke this endpoint in the Console UI:
...
+   "namespaces":["*"],
+   "appIDs":["*"],
+   "accountIDs":["*"],
+   "codeRepos":["*"],
+   "clusters":["*"], 
+   "color":"#AD3C21"
 }'
 ```
-Any resource left unspecified, such as `hosts`, `functions`, or `clusters`, is assigned a wildcard by default.
+**Note:** Resources that are not explicitly specified need to have a `*` wildcard applied to them in order to instantiate as mentioned in the example.
```
iansk: @ssugandh Let's remove this line. You're already adding a lot of info about this issue at the beginning of the endpoint description. No need to repeat it again here.

api/descriptions/collections/post.md
iansk: @ssugandh The reason I want our curl commands specified this way is because our OpenAPI spec doesn't declare which params are optional and which are required. As such, I want our curl commands to act as a stand-in for this missing info.

Customers can then take our "minimally useful" curl command, and build on it to implement their own use case.

api/descriptions/collections/post.md
ssugandh: @iansk : There's no response for a successful execution. I have retained the previous line with a caveat to verify with the Console UI.

api/descriptions/collections/post.md
ssugandh: @iansk : I've verified that only three resources are manadatory to use the collection, name, images, and labels. I have described about it in the description. I've retained the example with some resources just as a reference. Let me know if you still think otherwise, I can remove the non-mandatory resources from the example. 

In my other experiences to validate cURL example for creating users, I struggled to get the exact names of the property "role". I had to create users in the UI and use GET to understand all the exact values of the "role" used in the JSON file.

For color, should we not update about the random color in the parameter description?

ssugandh:
@iansk : Updated all.

kdpanw:
@iansk @Pubs-MV Hey team, did this ever get merged?

eric-bischoff (approved):
This is good to be merged.

Pubs-MV:
@ssugandh can you please confirm  that this does not conflict and can be merged?

ssugandh (approved):

ssugandh:
Yes, this can be merged. I don't have the merge permissions.

ssugandh:
@iansk and @Pubs-MV : Can you please merge this? I don't have merge permissions.

PR #124, opened 2021-12-06
Update audit_admin_activity.adoc

Pubs-MV:
replaced the URL: https://cdn.twistlock.com/docs/api/twistlock_api.html 
with https://prisma.pan.dev/api/cloud/cwpp/policies/


compute/admin_guide/audit/audit_admin_activity.adoc
```diff
@@ -13,15 +15,15 @@ Successfully authenticated calls to the API are not recorded.
...
 
 Settings and rule events show how a configuration has changed.
 Both the API endpoint, and a diff of the previous and current JSON objects are shown.
 The following screenshot shows the changes to a vulnerability rule:
 
 image::admin_activity_audit_trail_diffs.png[width=600]
 
-Use the https://cdn.twistlock.com/docs/api/twistlock_api.html[API reference] to determine what has changed and how to interpret the meaning of the change.
+Use the https://prisma.pan.dev/api/cloud/cwpp/policies/[API reference] to determine what has changed and how to interpret the meaning of the change.
 The `/api/v1/policies/cve` endpoint creates and modifies vulnerability rules.
```
Pubs-MV: @iansk this API endpoint seems dated. What should we update it to?

iansk: @Pubs-MV This part needs a bigger refactor. To fix immediately, delete everything starting from the new line 22 (`The following screenshot...`).

Alternatively, to refactor, we'd need to start with a new screenshot. The current one was taken back in 2017. The `/policies/cve` endpoint no longer exists, it's been replaced with dedicated `/policies/vulnerability/*` endpoints. Also, you can no longer use the API docs as a reference for the diffs, you'd need to use the downloadable OpenAPI spec, which covers all endpoints, not just versioned endpoints.

Pubs-MV: updated

iansk (approved):

PR #126, opened 2021-12-09
Mor shabi Autopilot defender instructions

MorShabiPAN:
Hi @iansk,

As discussed, I'm adding the following instructions for installing the *GKE Autopilot* defender.

I'll have to update the System requirements as well, but I'm still waiting for GKE to specify the versions.

We are linking to this documentation from a blog that will be published on **Dec 13 evening/Dec 14 morning by 6 am PT**, the blog itself is published on **Dec 14 Tue @9am PT**. It will be great if we could merge this PR at the same time as well.

Please let me know if anything else is needed.

Mor

cc @MayaShani


iansk:
@MorShabiPAN Merging this now. I'll publish it later today.

PR #127, opened 2021-12-09
Update book_prisma_cloud.yml

Pubs-MV:
Commented out the Upgrade announcement as these are in the RN, when applicable. This topic has a lot of dated content.


iansk (approved):

iansk:
Thanks @Pubs-MV. Merging this now

PR #128, opened 2021-12-09
Missing Incident Types in documentation

mallib-palo:
These three types are not in our public documentation:

Hijacked Process:
Data exfiltration:
Cloud Provider


iansk (approved):

iansk:
Thanks @Malik-PCC. I'm going to merge as-is, and do a little refactor on it. Content will go live a little later today

PR #174, opened 2022-02-01
Update install_openshift_3_11.adoc

timberbee:
Inside the documentation on line there are two Openshift entries in the TOC without a differentiator in the TOC heading. I am proposing we add 3.11 for this heading to distinguish it from the Openshift 4 installation page.


iansk:
This has been updated inline with your proposal already. Closing unmerged.

PR #195, opened 2022-02-16
Add new templates to illustrate markup

rcaballeromx:


template/concept-template.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+To include images, use the followinge markup to specify the path, filename, and the width of the image.
+
+[source]
+```
+image::screenshot-example.png[width=500]
+```
+
+Store the images under the `_graphics` folder of the pertinent guide.
+For example in the CNS SaaS guide, store images under `/sass\admin_guide/_graphics/`
```
Pubs-MV: typo in saas. Is there an issue with the / in the location path?

rcaballeromx: I would remove the backslash from the path since it can cause many issues. It doesn't seem to break the render here though.

Pubs-MV (approved):
Thanks @rcaballeromx

PR #212, opened 2022-02-28
Updates-FAQ

Pubs-MV:
added more


Pubs-MV:
some more clarification on CVEs for supported releases only

PR #217, opened 2022-03-01
Removed note for Executable File Access audit

grevach:
Removed "Processes that are known to write binaries, such as “cp” and “mv”, are excluded." from the description of the "Executable File Access" audit.

See https://spring.paloaltonetworks.com/twistlock/twistlock/issues/36585


Pubs-MV (approved):
Thank you, Gal!

PR #226, opened 2022-03-10
update log_rotation.adoc

kdpanw:
added annotation as the console log path is only available in self-hosted


kdpanw:
PR on behalf of Eric Bischoff - pcc tac

Pubs-MV (approved):
Thank you!

PR #233, opened 2022-03-15
Added note for internal fields

grevach:
Added a note explaining that the "image" object is for internal use and will return empty.


iansk (approved):

iansk:
Thanks @grevach. This is now merged.

PR #240, opened 2022-03-22
Updates for PCSUP-8574

Pubs-MV:
Clarification that we support JFrog on-prem; no support for JFrog Cloud


MorShabiPAN:
Hi @Pubs-MV it looks good to me, @MayaShani also approved these changes

PR #277, opened 2022-04-16
Add Detail to ECR Registry URL Step

jjchavanne:
It's important to get this syntax correct, otherwise scans do not work and unfortunately Prisma Cloud does not tell you this is wrong.  This additional detail will help ensure users enter correct URL format for this field.


Pubs-MV (approved):
Thank you for adding the format @jjchavanne

PR #279, opened 2022-04-20
Explicitly state what is dropped

jamcg:
Update to explicitly state compliance and vulnerability data is dropped when updating custom compliance rules.


iansk (approved):

iansk:
Thanks @jamcg. Merging this now

PR #286, opened 2022-04-25
Move syslog-forwarding.adoc

arifulhuq:
Moving this doc from reference to configure


arifulhuq:
@Pubs-MV @rcaballeromx please approve. Moving this under a more relevant topic.

PR #290, opened 2022-04-27
update defendertypes.adoc

kdpanw:
If Docker Engine is detected on the host, installation of this Defender type is blocked; install Container Defender instead. - this is not longer the functionality, you can install a host defender on a host with docker if you wish.

### DIFF
--- a/compute/admin_guide/install/defender_types.adoc
+++ b/compute/admin_guide/install/defender_types.adoc
@@ -47,7 +47,7 @@ Stand-alone Container Defenders are installed on hosts that are not part of a cl
 Host Defender utilizes Prisma Cloud's model-based approach for protecting hosts that do not run containers.
 This Defender type lets you extend Prisma Cloud to protect all the hosts in your environment, regardless of their purpose.
 Defender runs as a systemd service on Linux and a Windows service on Windows.
-
+If Docker is deployed on your host, deploy a container Defender to protect the containers and the underlying host.
 
 Deploy one Host Defender per host.
 Do not deploy Host Defender if you've already deployed Container Defender to a host.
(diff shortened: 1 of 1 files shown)


harisrini:
@kdpanw 


If Docker Engine is detected on the host, installation of this Defender type is blocked; install Container Defender instead.
1. Is this for the future feature? Today auto-deploy doesn't differentiate between vanilla hosts and container hosts, it deploys the host defenders everywhere. So if a user needs to deploy a container defender on such a host, they need to uninstall the host defender already deployed and then install the container defender
2.  In Kepler there is a feature  - Auto-deploy feature of host defenders will deploy host defenders only on hosts where there are no docker or Kubernetes utilities found. or in other words, it will skip container hosts
3. So what are you trying to say about this? Is it about 1 above?

cc @avish9

kdpanw:
@harisrini 

Hi Hari,
   This is not for auto-deploy or a future feature. This is for deploying a Single Container Defender OR a single Host defender.
   
   Our docs currently state that "If Docker Engine is detected on the host, installation of this Defender type(where the "Type" is a Host Defender) is blocked; install Container Defender instead. "
   
   - this is not how the product behaves. It will allow you to install whatever defender type you want, on a host with/without docker.
   
   Therefore, I think that we need to remove that line as it does not actually do a check to see whether or not Docker is installed on the machine.

harisrini:
That is correct, if there is a docker engine, that doesn't stop our host defenders from being deployed.
I recommend let's replace that with this - 
If there is a docker-engine deployed, rather than deploying the host defender switch to deploy the container defender for protecting both the containers and the underlying host.

rcaballeromx:
@kdpanw The key issue is that the product is not making the check that the line in the docs says it does, but we still need to discourage customers from deploying a host Defender. @harisrini Your suggestion to switch the line to a recommendation is a good one, and I have added a commit with the correct wording. PTAL and approve for publication.

kdpanw:
@rcaballeromx yes, I do agree. thank you team!

Pubs-MV (approved):
Thank you for the collaboration!

PR #305, opened 2022-05-04
Update how JFrog Artifactory registry scans work

iansk:
Update 22.01 update 2 release notes with additional details about how security scans work.

See twistlock/twistlock#37179


iansk:
@grevach Please review

compute/rn/release-information/release-notes-22-01-update2.adoc
```diff
@@ -59,10 +59,24 @@ Now PRs are properly created.
...
-// #32760
+// #32760, #37179
 * Updates how the Prisma Cloud Compute scanning process impacts artifact metadata in JFrog Artifactory.
 The scanning process no longer updates the *Last Downloaded* date for all manifest files of all the images in the registry.
 +
+This new capability is supported on JFrog Artifactory 7.21.3 and later.
+When configuring Prisma Cloud registry scan settings, the version must be set to *JFrog Artifactory*.
+This capability isn't supported when version is set to  *Docker v2*.
++
+This capability is only supported when you configure Prisma Cloud to scan https://www.jfrog.com/confluence/display/JFROG/Repository+Management#RepositoryManagement-LocalRepositories[local repositories].
```
grevach, line 70: Thanks @iansk 
The entire explanation is good, however, Jfrog scanning will work this way only starting Kepler
In Joule update 2, following the fix we made, the scanning for remote and virtual repos backed by remote repos will just fail.... virtual repos backed by local repos will work and won't update the last download.
In Kepler we must change to Docker APIs so the issues with remote repos won't happen, therefore the virtual repos backed by local repos will get their last download time updated

So I think this whole section should be in the release notes for Kepler. We can set it together with a part saying we will now not update the last download time of the images themselves (https://spring.paloaltonetworks.com/twistlock/twistlock/issues/37179)

grevach:
> @grevach Please review

Thanks @iansk, added my comment

Also - I think we should also explain it in the article for Jfrog registry scanning in Kepler, and not just in the release notes. What do you think?

iansk:
All right info is already in the  22.01 and 22.06 docs. Closing this unmerged.

PR #319, opened 2022-05-16
update-serverless.adoc

kdpanw:
AWS account permissions are not listed in compliance for function scanning . This can be confusing if for some reason a customer is setting up compliance checks but not vulnerability scanning.

### DIFF
--- a/compute/admin_guide/compliance/serverless.adoc
+++ b/compute/admin_guide/compliance/serverless.adoc
@@ -158,38 +158,7 @@ If you wish to add an account, click on *Add credential*.
  
 .. Click *Add*.
 
-=== Authenticating with AWS
-
-The serverless scanner is implemented as part of Console.
-The scanner requires the following permissions policy:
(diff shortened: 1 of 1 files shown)


kdpanw:
@Pubs-MV 

This might be better as a blurb with an xref to : https://docs.paloaltonetworks.com/prisma/prisma-cloud/prisma-cloud-admin-compute/vulnerability_management/serverless_functions | you make the judgement call though. regardless, the permissions should be mentioned somewhere a customer can see in the doc for serverless compliance checks, because it won't work without the account having that setup.

-best

Pubs-MV:
Thanks for the recommendation @kdpanw

Pubs-MV (approved):
@kdpanw , I edited the workflow to add a step with a link to the topic with the required permissions. 
The syntax does not allow for adding a heading under a set of steps, so modified your updates. Thanks for chipping in!

PR #326, opened 2022-05-19
update-install_amazon_ecs.adoc

kdpanw:
--non related to case or jira, change to correct consistency and reflect the actual values in the json, see : https://cdn.twistlock.com/docs/attachments/amazon-ecs-task-pc-console.json


Pubs-MV (approved):
Thanks @kdpanw

PR #330, opened 2022-05-23
Update Kubernetes auditing doc

ytzur:
- Fix typographical and grammatical errors
- Remove references to old versions of Kubernetes and GKE that are no longer officially supported
- Remove suggestion to disable and enable Kubernetes auditing to immediately get new audits which as of writing this is not accurate (it causes the Console to immediately query the cloud provider but the time range used in the query targets logs from five minutes before, might not include recent logs).
- Clarify guidance on exporting Kubernetes audits in AKS and EKS


rcaballeromx (approved):
Thanks for making those fixes

PR #332, opened 2022-05-24
added UI path to disable automatic learning

Pubs-MV:


grevach (approved):

Pubs-MV:
Update from Usabilla feedback

PR #335, opened 2022-05-25
Fix IAM role for copy and paste

SimOnPanw:
Comma was the wrong place


Pubs-MV (approved):
Thank you, @SimOnPanw

PR #341, opened 2022-06-02
Added setup of external route for Defenders comms

azaceta:
I found the current instructions for OpenShift 4 didn't allow the communication of external Defenders back to the Console, so I:
Added the definition of an external route to expose port 8084 to enable communication of Defenders to the Console.
Added setup of the route's URL in the SAN of the console, and in the Defender installation instructions, also adding the option to specify the target port from the external Defender point of view.


wfg (approved):
@azaceta Thanks for the addition. LGTM!

@pfox1969 one more set of eyes please?

wfg:
@Pubs-MV good to go 👍

PR #343, opened 2022-06-02
update_serverless_scanning.adoc

kdpanw:
non-jira from pcs-se-twistlock, sean s , gary K


kdpanw:
@Pubs-MV  this is invalid - do not merge

PR #344, opened 2022-06-03
Update custom_certs_predefined_dir.adoc

lstpierre-panw:
Removed the ")." from the end of the command on line 150.


Pubs-MV (approved):
Thank you @lstpierre-panw

PR #349, opened 2022-06-08
adding RN

Pubs-MV:


compute/rn/release-information/release-notes-22-01-update-3.adoc
```diff
@@ -0,0 +1,112 @@
...
+|22.01.889
+
+|Code name
+|Joule, 22.01
+
+|Release date
+|June xx, 2022
+
+|Type
+|Maintenance release
```
Pubs-MV: Is this a maintenance release?

iansk: Yes, we've been calling these maintenance releases.

Also, since it's limited to fixes for vulns, we've been calling them re-issues of the latest release. So, in this case, this would be a re-issue of update 2, and we'd uniquely identify which "update 2" it is by its build number.

iansk (approved):

PR #356, opened 2022-06-09
Addition of deprecation for Windows Server 2016

Pubs-MV:


compute/rn/release-information/release-notes-22-06.adoc
```diff
@@ -647,6 +647,10 @@ Fargate tasks protected by App-Embedded Defender will be reported under the *App
 // #34188
 * Support for Openshift 3.11 will be removed in the next release, Lagrange.
 
+// # 39473 (email from Mor)
+* Support for Windows Server 2022 will be added in Lagrange, and Windows Server 2016 will no longer be supported.
```
MorShabiPAN: Hi @PaloAltoNetworks/announcements we would prefer to say by Lagrange since it might happen before Lagrange. We will stop supporting WS16 when we will add WS22.

Pubs-MV: got it!

Can I update to say,

`* Support for Windows Server 2022 will be added with or before the next release, Lagrange. With support for Windows Server 2022, Windows Server 2016 will no longer be supported.`

MorShabiPAN: @PaloAltoNetworks/announcements sounds good.

rcaballeromx (approved):
Looks good to me.  I checked the links and both point to the correct pages.

PR #367, opened 2022-06-13
Improve heading in System Requirements

rcaballeromx:


tallieber:
I think maybe the title should be "ARM architecture-specific Requirements" so that the "Architecture" wording would not confuse at  first sight 
@MayaShani @rcaballeromx

tallieber (approved):

compute/admin_guide/install/system_requirements.adoc
```diff
@@ -81,7 +81,7 @@ Prisma Cloud has been tested on the following services:
 * Microsoft Azure
 * Oracle Cloud Infrastructure (OCI)
 
-=== Architecture: ARM64
+=== ARM Architecture-specific Requirements
```
Pubs-MV: Do we need the word specific? Would calling it ARM Architecture Requirements work?
If yes please make it Specific with "S".

And for the lead in sentence, please consider replacing 
To use Prisma Cloud on ARM architecture, `consider ` the following requirements with `adhere to` or `review`

tallieber: we dont need the "specific" word, thanks for the suggestion. 
about the "review" or "consider" I will let @rcaballeromx decide, to be consistent with all the doc

Pubs-MV (approved):

PR #370, opened 2022-06-14
Update system_requirements.adoc

tallieber:
added tas support v2.13


Pubs-MV (approved):

Pubs-MV:
change per #38764 (on spring)

PR #380, opened 2022-06-21
Update kubernetes_auditing.adoc

MorShabiPAN:
Auto-completion doesn't work in K8s audit rules.

@MayaShani @Pubs-MV, can you please review this suggestion?


Pubs-MV (commented):
Hey @MorShabiPAN , is this capability `Press `OPTION` + `SPACE` to get a list of valid terms, expressions, operators, etc, for the given position.not supported` or did you remove it because the buttons are specific to a Mac?

Pubs-MV (approved):

PR #401, opened 2022-07-07
[Usabilla feedback] Update crypto_miners.adoc

Pubs-MV:
feedback from Usabilla


pfox1969 (approved):
clarification on the fact the automatic runtime learning needs to be enabled

PR #416, opened 2022-07-14
[Template]PCSUP#23144 Update for inline image

Pubs-MV:


Pubs-MV (commented):
fixed the syntax for inline image

PR #425, opened 2022-07-17
adding minimum defender versions and the new req.host functionality 

e-shuster:


compute/admin_guide/waas/waas_custom_rules.adoc
```diff
@@ -92,81 +92,108 @@ For example:
...
-|req.headers
+|req.host
+|22.06
+|Map of String
+|`req.host contains /^.*ACME[1-9]{1,6}$/`
+
+|req.headers +
+
+(for matching on "Host" header use req.host)
+|21.04
```
Pubs-MV, line 110: @e-shuster , 21.04 is EoL. I do not think we should keep an EoL version as the minimum version. Not sure what we have done in the past.

e-shuster: @Pubs-MV - this just states that this syntax will be supported from this version on. I don't think n-2 backwards compatibility is an issue here. This column refers to the minimum version.

Pubs-MV: I should clarify @e-shuster. I meant to say that we should not mention an EoL version as the minimum supported version. Merging it in now, but we should talk about this to be clear.

Pubs-MV (approved):

PR #431, opened 2022-07-19
Add an example for AWS/GCP

rdsingh11:
Can you enhance this doc and add the steps for AWS/GCP


Pubs-MV:
Hi @rdsingh11, if we move the public cloud information from heer: https://docs.paloaltonetworks.com/prisma/prisma-cloud/22-06/prisma-cloud-compute-edition-admin/authentication/credentials_store to this topic, will the requirement be covered?

Pubs-MV:
addressed as part of the Lagrange reorg of content

PR #441, opened 2022-07-21
[RN22.06] Update to clarify deprecation notice

Pubs-MV:
#36042; CWP will not deprecating CI pipeline vulnerability scanning with twistcli only repo-scanning.


Pubs-MV:
Reviewed on Slack with John M and Guy E
cc: @MayaShani

PR #452, opened 2022-07-26
Added required permission

avish9:


Pubs-MV (commented):
is there a GH issue or PCSUP that you are fixing here @avish9 . please reference it, if possible

PR #467, opened 2022-07-30
Agentless configuration edits

Aqsaxvv092:
- Removed repeatitive / unnecessary content
- Created easy to follow steps with concise actions for users to follow
- Explained each scanning type which was missing
- Added images where needed
- Added default configuration fields section for SaaS users who will see the note from UI.
- Cleaned up entire flow end to end.


compute/admin_guide/configure/configure-agentless-scanning.adoc
```diff
@@ -50,73 +31,74 @@ ifdef::prisma_cloud[]
...
 . Click on *Add Account* or click the edit icon for an existing account.
 +
 image::add-account.png[width=400]
 
-. Select your cloud provider, configure its https://docs.paloaltonetworks.com/prisma/prisma-cloud/22-06/prisma-cloud-compute-edition-admin/authentication/credentials_store[credentials], and click *Next*.
+. Select your cloud provider, add its https://docs.paloaltonetworks.com/prisma/prisma-cloud/22-06/prisma-cloud-compute-edition-admin/authentication/credentials_store[credential].
 +
 image::set-credentials.png[width=800]
++
+NOTE: You can leave permissions empty while creating these credentials. Prisma Cloud will provide Agentless permission templates in the next step after validating the given credentials. For incorrect credentials, permissions template download will show an error.
```
rcaballeromx: @Aqsaxvv092 The behavior in this note isn't clear. What fields can be left empty? Do you mean the fields listed in lines 55-57 for each provider?

Pubs-MV: @Aqsaxvv092 we are merging this PR now, but we need an answer from you to resolve this question.

Aqsaxvv092: Users who create credentials in public clouds like AWS. This is something they do on cloud side not on our side. While creating credentials in AWS there is a step in which they can provide permissions to the credential / token. They need to just leave it empty during that stage.

PR #477, opened 2022-08-02
Add steps to integrate Azure AD and OIDC

### LINKED ISSUE #448: Document OpenID Connect support for Azure AD
We currently have an article describing PCC integration with OpenID Connect: https://docs.paloaltonetworks.com/prisma/prisma-cloud/22-06/prisma-cloud-compute-edition-admin/authentication/oidc.

The article only describes the steps for PingOne and Okta, and does not refer to the process required in Azure AD, which is also supported with OpenID Connect.

We should extend the article to include the required steps for integrating with Azure AD.

The request also came up in this support case from Morgan Stanley: https://redlock.atlassian.net/browse/PCSUP-10521

rcaballeromx:


compute/admin_guide/authentication/oidc.adoc
```diff
@@ -87,7 +87,6 @@ Enter the following callback URL:
 +
 *\https://<CONSOLE>:<PORT>/api/v1/authenticate/callback/oidc*, and then click *Save*.
 
-
 [.task]
 ==== Configure Prisma Cloud
```
grevach: @rcaballeromx this is also configuration of Prisma Console, so it should be somehow connected to the configuration of the Console for Azure. (since this section also shows how to create the user / group in the console and then make sure you can connect with it to the Prisma Console)

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -116,12 +115,92 @@ Configure Prisma Cloud.
...
+
+. Enable OpenID Connect.
+
+. Enter the following information in the settings fields.
+
+.. *Client ID*: The Application ID for the client you have from Azure AD.
+.. *Client Secret*: The secret for the client that you created for the application and stored safely for later use.
+.. *Issuer URL*: The endpoint of the application registered in Azure AD, for example https://login.microsoftonline.com/`<TENANT_ID>`/v2.0
+.. *Group scope*: Leave this field blank.
+.. *Group claim*: The name of the group you created in Azure AD.
```
grevach: @rcaballeromx I believe this should be "groups" and not "The names of the group you created in Azure AD"

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -116,12 +115,92 @@ Configure Prisma Cloud.
...
+. To add the secret for the client, go to *certificates & secrets*.
+
+. Add a new secret for the client, copy and store it for later use.
++
+[IMPORTANT]
+====
+You can only view the value of the secret when you create it. Copy and store the secret safely for later use.
+====
+
+. To add the needed claim, go to *Token Configuration*.
```
grevach, line 165: @rcaballeromx I think this should be divided to a different section, this is quite a long procedure that I think that might be hard to track

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -116,12 +115,92 @@ Configure Prisma Cloud.
...
+
+. Add a group claim and select the groups you want to assign to the application.
+
+. To create the needed application group, go to *Groups* in the Azure AD console.
+
+. Create a new group and keep the default values.
+
+. To assign the newly created group to the Prisma Cloud Console, go to *Enterprise applications* in the Azure AD console.
```
grevach: @rcaballeromx maybe this can be divided as well

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -116,12 +115,92 @@ Configure Prisma Cloud.
...
+[.task]
+=== Azure Active Directory (AD)
+
+To integrate with Azure Active Directory (AD), you must register Prisma Cloud as an Open ID Connect (OIDC) application in Azure and configure Azure AD as an identity provider in Prisma Cloud.
+
+[.procedure]
+
+. Go to https://portal.azure.com/#home[your Azure console].
+
+. Find the Azure AD service.
```
Tomerit, line 142: Azure AD -> Azure Active Directory

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -116,12 +115,92 @@ Configure Prisma Cloud.
...
+
+. Click the *app registration* button and select *New registration*
+
+. Enter a name and select *Twistlock only - single tenant* as the supported account type.
```
Tomerit: That is based on Customer env (drop Twistlock only - single tenant). Can direct customers to choose - 'Accounts in this organizational directory only', but as mentioned that’s env dependent.

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -116,12 +115,92 @@ Configure Prisma Cloud.
...
+
+. Under *Redirect URI* select *Web platform* and enter the path to the Prisma Cloud Console API for OIDC, for example https://`<CONSOLE_URL>`:8083/api/v1/authenticate/callback/oidc
```
Tomerit: Web platform -> Rephrased by Azure to ‘Web'
CONSOLE_URL -> not consistent with previous convention in other providers

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -116,12 +115,92 @@ Configure Prisma Cloud.
...
+. Add a new secret for the client, copy and store it for later use.
++
+[IMPORTANT]
+====
+You can only view the value of the secret when you create it. Copy and store the secret safely for later use.
+====
+
+. To add the needed claim, go to *Token Configuration*.
+
+. Add a group claim and select the groups you want to assign to the application.
```
Tomerit: select the groups you want to assign to the application -> 	select ‘groups assigned to the application’. This is a general settings.

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -116,12 +115,92 @@ Configure Prisma Cloud.
...
+
+. Click add and select your user.
+
+. Go to *App registrations* in the Azure AD console.
+
+. Click on *Your owned registered app*.
+
+. Find the application you registered and click on *Endpoints*.
+
+. Find and download the OpenID Connect metadata JSON file.
```
Tomerit: no need to download can open in browser

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -116,12 +115,92 @@ Configure Prisma Cloud.
...
+. Copy the value under Issuer URL from the JSON file, for example: https://login.microsoftonline.com/`<TENANT_ID>`/v2.0
+
+[.task]
+==== Configure Azure AD as an Identity Provider
+
+After you register Prisma Cloud as an Open ID Connect (OIDC) application in Azure, complete the following steps to configure Azure AD as an identity provider.
+
+[.procedure]
+
+. Go to *Manage > Authentication > Identity Providers > OpenID Connect* in your Prisma Cloud Console.
```
Tomerit: drop 'OpenID Connect' from nav

compute/admin_guide/authentication/oidc.adoc
Tomerit: +1

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -87,7 +87,6 @@ Enter the following callback URL:
 +
```
Tomerit, line 99: - Line 3 - an standard -> a standard
    - Line 5 - Add Azure AD

grevach:
Thanks @rcaballeromx, please see Tomer's and my comments

compute/admin_guide/authentication/oidc.adoc
rcaballeromx: @grevach @Tomerit Is "groups" an option they select from a dropdown menu? If it's not, the label alone is not clear enough without further context. What are the "groups" that customers need to enter here? Where can they find them? Maybe what we need to include here is an example to make it clear to customers what information they need to enter.

compute/admin_guide/authentication/oidc.adoc
rcaballeromx: @grevach I agree 100% that this procedure is too long and I am adding sections where you suggested. 
That said, adding sections doesn't solve the issue that it's difficult to track and complete for customers.
This procedure should be simplified to make the customer experience more streamlined.

compute/admin_guide/authentication/oidc.adoc
rcaballeromx: Thank you for clarifying. @Tomerit I'm calling out explicitly that it's a setting customers need to select.

compute/admin_guide/authentication/oidc.adoc
rcaballeromx: @Tomerit Thank you for letting me know, but I don't understand from the comment what the new phrasings should be:
Should it be **Web console URL** instead of **Web platform**?
Should it be `<WEB_CONSOLE_URL>` instead of `<CONSOLE_URL>`?

compute/admin_guide/authentication/oidc.adoc
rcaballeromx: What exactly do you mean with it being environment dependent?
What conditions need to be met for a customer to choose **Twistlock only - single tenant** over **Accounts in this organizational directory only**? 
Are there other options? Which is the most commonly used by our customers?
Aside from that, my question was also regarding the use of "Twistlock". I understand that may be how Azure has it listed but shouldn't it be instead "Prisma Cloud Compute"?
We'll document it as it appears in the UI for customers, but I mention it because we are trying to reduce the use of Twistlock in our docs in favor of Prisma Cloud Compute.

compute/admin_guide/authentication/oidc.adoc
rcaballeromx: @grevach No, the section starting in line 91 is a subsection for the configuration of the Okta integration.

The Azure AD equivalent starts in line 177. 

Each integration has it's own section where customers create the users or groups needed for that integration to work.

To avoid confusion, I see two alternatives:

* Change the heading in line 91 to `==== Configure Okta as an Identity Provider` OR
* Change the heading in line 177 and the heading in line 91 to `==== Configure the Prisma Cloud Console`

The better alternative is the first because it provides better SEO. When the page is indexed by search engines, headings heavily influence the quality of search results and the heading "Configure the Prisma Cloud Console" won't be as effective. WDYT?

compute/admin_guide/authentication/oidc.adoc
rcaballeromx: @Tomerit We don't need to spell out the abbreviation here because it has been spelled out on first mention just above.

compute/admin_guide/authentication/oidc.adoc
rcaballeromx: I've added a list of all providers with in-page links to make it easier for customers to jump to the instructions they need.

compute/admin_guide/authentication/oidc.adoc
Tomerit: 'groups' is the groups claim (key) name.
Its value is only the groups that are explicitly assigned to the application and that the user is a member of (if 'Groups assigned to the application' has selected), which is recommended for large organizations due to the group number limit in token.
The value is auto populated in the token, we recognize it by the groups claim key name.

compute/admin_guide/authentication/oidc.adoc
Tomerit: 'Twistlock' is our organizational directory name, and irrelevant in a customer env context.
Replace with - 'Accounts in this organizational directory only'

compute/admin_guide/authentication/oidc.adoc
Tomerit: https://\<CONSOLE\>:\<PORT\>/api/v1/authenticate/callback/oidc
Like in PingOne and Okta

compute/admin_guide/authentication/oidc.adoc
grevach: @rcaballeromx ok, if this is the case then I agree using `==== Configure Okta as an Identity Provider` and `==== Configure Azure AD as an Identity Provider` is better.
However, I don't see an equivalent section for PingOne which is also one of the supported providers.

compute/admin_guide/authentication/oidc.adoc
grevach: @rcaballeromx well, the procedure depends on the configuration at Azure's side, so at the moment I can't see a way to simplify it. If something changes in the future we'll of course adjust it

compute/admin_guide/authentication/oidc.adoc
rcaballeromx: Added the context to the doc. Thanks @Tomerit

compute/admin_guide/authentication/oidc.adoc
rcaballeromx: @grevach I've made the change to the headings. As far as I could tell, the PingOne section only includes the configuration changes that take place on the PingOne side, and that's why it doesn't have an equivalent section. @Tomerit Is that observation correct?

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -116,12 +124,101 @@ Configure Prisma Cloud.
...
+
+. Go to https://portal.azure.com/#home[your Azure console].
+
+. Find the Azure AD service.
+
+. Click the *app registration* button and select *New registration*
+
+. Enter a name and select *Accounts in this organizational directory only* as the supported account type.
+
+. Under *Redirect URI* select *Web console URL* and enter the path to the Prisma Cloud Console API for OIDC, for example *\https://<WEB_CONSOLE_URL>:8083/api/v1/authenticate/callback/oidc*
```
rcaballeromx: We can't recommend the use of non-versioned APIs in our docs. Do we have an alternative for customers?

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -56,13 +65,13 @@ Update the callback URL.
...
-
+[#okta]
 === Okta
 
 Integrate with Okta.
 
-* Initiate Login URI (Okta) must point to \https://<CONSOLE>:<PORT>/callback.
-* Redirect URI must point to \https://<CONSOLE>:<PORT>/api/v1/authenticate/callback/oidc.
+* Initiate Login URI (Okta) must point to *\https://<CONSOLE>:<PORT>/callback*.
+* Redirect URI must point to *\https://<CONSOLE>:<PORT>/api/v1/authenticate/callback/oidc*.
```
rcaballeromx: The Okta integration is also using the non-versioned API.

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -1,25 +1,34 @@
...
 You need to configure Compute as an OIDC app.
 When configuring your app:
 
-* The Start SSO URL must point to \https://<CONSOLE>:<PORT>/callback.
-* The Redirect URI must point to \https://<CONSOLE>:<PORT>/api/v1/authenticate/callback/oidc.
-* UserInfo must include sub, idpid, name.
-* All of the following scopes must be included for OpenID: OpenID Connect (openid), OpenID profile, OpenID Email, OpenID address, OpenID Phone, Groups. 
-
+* The Start SSO URL must point to *\https://<CONSOLE>:<PORT>/callback*.
+* The Redirect URI must point to *\https://<CONSOLE>:<PORT>/api/v1/authenticate/callback/oidc*.
```
rcaballeromx: The PingOne integration is also using the non-versioned API.

rcaballeromx (commented):
@grevach @Tomerit I've made all the changes you suggested. Thanks for reviewing.
Unfortunately, I noticed that all three integrations are using the `https://<CONSOLE>:<PORT>/api/v1/authenticate/callback/oidc` non-versioned API. I confirmed that the `v1/authenticate` API is not supported and thus we shouldn't recommend its use to customers. Do we have an alternative path for enablement of these integrations?

compute/admin_guide/authentication/oidc.adoc
grevach: @rcaballeromx I don't believe we do... @Tomerit are you familiar with any other way the customer can configure the integration without using this API (https://<WEB_CONSOLE_URL>:8083/api/v1/authenticate/callback/oidc)?

In any case, I'll add this API to the GH issue for adding supported APIs in Lagrange

grevach:
> @grevach @Tomerit I've made all the changes you suggested. Thanks for reviewing. Unfortunately, I noticed that all three integrations are using the `https://<CONSOLE>:<PORT>/api/v1/authenticate/callback/oidc` non-versioned API. I confirmed that the `v1/authenticate` API is not supported and thus we shouldn't recommend its use to customers. Do we have an alternative path for enablement of these integrations?

@Tomerit are you familiar with any other workaround to configure the integrations?

compute/admin_guide/authentication/oidc.adoc
```diff
@@ -199,9 +206,9 @@ After you register Prisma Cloud as an Open ID Connect (OIDC) application in Azur
 
 .. *Client ID*: The Application ID for the client you have from Azure AD.
 .. *Client Secret*: The secret for the client that you created for the application and stored safely for later use.
-.. *Issuer URL*: The endpoint of the application registered in Azure AD, for example https://login.microsoftonline.com/`<TENANT_ID>`/v2.0
+.. *Issuer URL*: The endpoint of the application registered in Azure AD, for example *\https://login.microsoftonline.com/<TENANT_ID>/v2.0*
 .. *Group scope*: Leave this field blank.
-.. *Group claim*: The name of the group you created in Azure AD.
+.. *Group claim*: These are the groups explicitly assigned to the application that include the user as a member. The value is populated automatically.
```
Tomerit: It's still not clear that we need the group claim key name - 'groups' here.
The value is populated automatically, but we must know the key name which is usually 'groups'

compute/admin_guide/authentication/oidc.adoc
Tomerit: I'm not sure I follow why is this route problematic

rcaballeromx: @Tomerit The issue is that the `authenticate/callback/oidc` API is not supported. You can review the versioned and supported options of the [API in our docs](https://prisma.pan.dev/api/cloud/cwpp/authenticate).
That's why @grevach is adding it to the GH issue to make it part of the versioned API.

grevach:
> > @grevach @Tomerit I've made all the changes you suggested. Thanks for reviewing. Unfortunately, I noticed that all three integrations are using the `https://<CONSOLE>:<PORT>/api/v1/authenticate/callback/oidc` non-versioned API. I confirmed that the `v1/authenticate` API is not supported and thus we shouldn't recommend its use to customers. Do we have an alternative path for enablement of these integrations?
> 
> @Tomerit are you familiar with any other workaround to configure the integrations?

@rcaballeromx I talked to Tomer, there is no other way to configure the integration without this API, I'll think about possible solutions and check with R&D and I'll get back to you

grevach:
@rcaballeromx @Pubs-MV 
As discussed we'll be adding the required API to the docs with an exception, and then publish this article
https://spring.paloaltonetworks.com/twistlock/twistlock/issues/40680

compute/admin_guide/authentication/oidc.adoc
grevach: @rcaballeromx please fix this as Tomer explained. What the customer needs to set here is the key, which is just explicitly 'groups', without any specific group name, these are populated automatically.

rcaballeromx: @grevach I rephrased to emphasize the value customers need to use.

rcaballeromx:
@grevach I've added the fact that the API is not included in our reference guide since it's only a configuration value.

Pubs-MV (approved):

PR #485, opened 2022-08-08
[Compute docs]: Fixing toc and UI paths in Vulnerability Explorer docs

divyabhushan-okta:
## Description

Added TOC for sub-folders. Re-worked on the headings and updated the new UI paths.

### DIFF
--- a/compute/admin_guide/vulnerability_management/customize_image_scanning.adoc
+++ b/compute/admin_guide/vulnerability_management/customize_image_scanning.adoc
@@ -14,7 +14,7 @@ Note that Prisma Cloud maps the Common Vulnerability Scoring System (CVSS) to a
 xref:../vulnerability_management/cvss_scoring.adoc#[grading system that ranges from Low to Critical.]
 
 [.procedure]
-. Open Console, and go to *Defend > Vulnerabilities > Policy*.
+. Open Console, and go to *Defend > Vulnerabilities > Images > Deployed > Add rule*.
 
 . Click *Add rule*.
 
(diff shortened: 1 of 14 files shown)


Pubs-MV (approved):
Thank you @divyabhushan

Pubs-MV:
hey @iansk can you please confirm that the force push is ok before I merge the changes

iansk:
@Pubs-MV Yes, this is safe to merge.

iansk:
@divyabhushan (cc: @Pubs-MV)

Hi Divya - In general, you should never force push to the repo, even it's in your own branch. If you find that you're stuck, and your only option appears to be a force push, don't hesitate to ping me for help.

In this case, I see that you wanted to tweak your original commit with a small fix, so you squashed the "fix" commit onto the original commit, which gave you a brand new commit, completely replacing the original commit. Git wouldn't let you push this new commit without the `--force` option. That's because your new commit rewrites history. Essentially, your original commit disappears for everybody, and it's replaced with this new commit.

Instead, it's safer to keep your second fix commit, and push it to the repo as well. Your branch history would then show two commits: your original commit and the fix commit. The history in your branch may appear subjectively "messier", but it doesn't matter for the master branch, which is where we want to keep history clean. When we merge PRs into master, we squash all commits in your branch into a single commit with a single message.

Squashing commits with `rebase -i` is an acceptable practice, but only do it before pushing work in your branch to GitHub. Once it's pushed to GitHub, it becomes part of the public record, and it  shouldn't be changed.

If you've got any questions, just let me know! :)

PR #496, opened 2022-08-12
Removed a step for PCEE

Pubs-MV:
Applied the condition for the step to Click the Download button to get template files and reference to permission templates document that is applicable only for PCCE.


Pubs-MV:
Note that the link to the permission template doc was also a 404 because the permission templates are part of the on-prem PCCE book file only

Pubs-MV:
Aqsa approved on Slack

PR #502, opened 2022-08-17
[Compute AG] Fix for PCSUP-10093 and render issue

iansk:
Move content out of the admonition because the content isn't rendering properly on the HTML site (it's rendering fine in the PDF).


compute/admin_guide/continuous_integration/jenkins_pipeline_project.adoc
```diff
@@ -73,10 +70,15 @@ Note that the previous scan step already published the results in Console, and t
 
 .. In the *Sample Step* drop-down, select *prismaCloudPublish - Publish Prisma Cloud analysis results*.
 
-.. In *Scan Result Files*, accept the default.
+.. In *Scan Result Files*, review the json filename.
++
+If you have configured scanning for multiple images and configured unique filenames for each scan in the previous step, you must add a wildcard to the json filename for scan results. For example `prisma-cloud-scan-results*.json`. This ensures that the results of the scan are not overwritten and you can view it in the Jenkins plugin.
```
MorShabiPAN: @Pubs-MV, maybe we can change it from "This ensures that the results of the scan are not overwritten and you can view it in the Jenkins plugin." to something like:
"This ensures that the publish command will read all the result files with the same name pattern, and publish them"

Pubs-MV: @MorShabiPAN done. Please approve for publish.

MorShabiPAN (approved):

PR #503, opened 2022-08-17
[Compute API] API issue # 40773 - Remove unsupported endpoints and update missing titles

ssugandh:
- GH# 40773 - Removed the endpoint that downloads the MMDB file from DB Lite database.
- Added titles to the /util endpoints.
- Corrected Defenders serverless cURL example.
- Updated the topic map file.


Pubs-MV:
@ssugandh , per https://spring.paloaltonetworks.com/twistlock/twistlock/issues/40773, we should unpublish this endpoint. PTAL and let @MayaShani and  me know if you have questions.

ssugandh:
@Pubs-MV : I have removed other unsupported and v1 endpoints also from our topic map.

Pubs-MV (approved):
please confirm that these endpoints are marked as supported but not b/c, if that was agreed with @MayaShani 

approving to ungate you, @ssugandh

ssugandh:
@Pubs-MV and @iansk : Please see if this can be merged. I don't have the merge permissions.
This will close the issue here: https://spring.paloaltonetworks.com/twistlock/twistlock/issues/40773

PR #508, opened 2022-08-18
Update to match current console

sgordon46:
update paths as they have updated in console.


sgordon46:
@Pubs-MV

Pubs-MV (approved):
Thank you @sgordon46

PR #509, opened 2022-08-18
Update for app.jp stack

Pubs-MV:


wfg (approved):
I double checked GCP and swapped the two addresses. Looks good to me.

PR #515, opened 2022-08-25
[Compute AG] Fix #457: ATTACK Explorer - Map audits to techniques

### LINKED ISSUE #457: Add the audit types to ATTACK explorer techniques mapping to the docs
In one of the previous releases, we added the ATTACK explorer.
Each of our audits (runtime, admission, etc.) is mapped to single or multiple techniques.

We should add this mapping to the ATTACK explorer article https://docs.paloaltonetworks.com/prisma/prisma-cloud/22-06/prisma-cloud-compute-edition-admin/runtime_defense/attack.

The mapping is in this spreadsheet: https://docs.google.com/spreadsheets/d/1aWuAOj3eSwBVTWadw_QGDvZRBgtjOdTXz1Ih6ptHgus/edit#gid=1878487739.

In the spreadsheet:
* **Techniques** - are the techniques that the audit is always mapped to
* **Possible Additional Techniques** - are techniques that the audit can optionally be mapped to, depending on changing information from the audit. For example, for some audits, on new files being created we will check if the process that created the file is a compiler, and if so, we also map the audit to the Compile After Delivery technique.

iansk:


Pubs-MV (approved):
Looks good, thank you!

PR #518, opened 2022-08-26
[37516] PCSUP-8609 Console on Fargate updated based upon testing results.

pfox1969:
Update to Fargate Console install and backup / restore procedures.


Pubs-MV (approved):
Thank you @pfox1969

PR #519, opened 2022-08-26
[Compute AG]-[PCSUP-10975]

Pubs-MV:
Clarification that the the ‘project’ field that you see under the ci scan results will always present the jenkins job name.

### DIFF
--- a/compute/admin_guide/continuous_integration/jenkins_pipeline_project.adoc
+++ b/compute/admin_guide/continuous_integration/jenkins_pipeline_project.adoc
@@ -190,7 +190,7 @@ pipeline {
                 key: '',
                 logLevel: 'info',
                 podmanPath: '',
-                // The project field below is only applicable if you are using Prisma Cloud Compute Edition and have set up projects (multiple consoles) on Prisma Cloud. This field is not the name of the Jenkins pipeline.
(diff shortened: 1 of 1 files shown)


tallieber:
I think that it still misleading, because this field in the request can be a value of a tenant project. Regarding the “project” column in the scan result view we always display the Jenkins pipeline name

compute/admin_guide/continuous_integration/jenkins_pipeline_project.adoc
```diff
@@ -188,6 +190,7 @@ pipeline {
                 key: '',
                 logLevel: 'info',
                 podmanPath: '',
+                // The project field below is only applicable if you are using Prisma Cloud Compute Edition and have set up projects (multiple consoles) on Prisma Cloud. This field is not the name of the Jenkins pipeline.
```
tallieber: I think this is enough "The project field below is only applicable if you are using Prisma Cloud Compute Edition and have set up projects (multiple consoles) on Prisma Cloud." The second sentence will just confused them

Pubs-MV:
Thank you @tallieber . Can you please approve to merge?

tallieber (approved):

PR #530, opened 2022-09-02
Add DNS note to self-hosted servers

mikeurbanski1:


Pubs-MV (approved):
thank you @mikeurbanski1

PR #535, opened 2022-09-07
update-helm-install-changes

kdpanw:
--create-namespace | flag changed in recent helm updates, see this for reference: https://helm.sh/docs/helm/helm_install/


kdpanw:
See v2 where --namespace used to be an applicable flag 
https://v2.helm.sh/docs/helm/#helm-create

Pubs-MV (approved):

PR #544, opened 2022-09-13
[Compute Admin] Update twistcli_scan_images.adoc

kdpanw:
Twistcli --tarball utilizes chroot, which is a sudo or root user level operation in Linux . Therefore, we need to specify this so customers will know to expect this behavior . 

@mklambert for contributions to this PR .

Non-PCSUP


kdpanw:
@Pubs-MV can you make sure my edits do not break the table / note ? I'm not the best at making sure the formatting is preserved .

Pubs-MV:
also includes the update for [PCSUP-11309](https://redlock.atlassian.net/browse/PCSUP-11309) and https://redlock.atlassian.net/browse/PCC-1969

PR #550, opened 2022-09-16
refactor agentless configuration

tkishel:
This potentially improves our Configure Agentless Scanning documentation. 

It ...

- Consolidates overlapping documentation from two separate pages into one (agentless-permission-templates.adoc into configure-agentless-scanning.adoc).
- Highlights configuring the Hub account first, which is required when using Hub Account Scanning Mode for a Target.
- Includes adding credentials when using SaaS, which was omitted in favor of reusing credentials, but both are valid options.
- Refers to GCP Projects as Projects instead of Accounts.

Note that I cannot verify that I created the internal link to `Scanning Modes` correctly.

Also note, the `configure-agentless-scanning` page should have Version navigation (SaaS, 22-06) like this page does:

https://docs.paloaltonetworks.com/prisma/prisma-cloud/prisma-cloud-admin-compute/configure/certificates

Yet it does not:

https://docs.paloaltonetworks.com/prisma/prisma-cloud/prisma-cloud-admin-compute/configure/configure-agentless-scanning

https://docs.paloaltonetworks.com/prisma/prisma-cloud/22-06/prisma-cloud-compute-edition-admin/configure/configure-agentless-scanning

But I do not know how to enable/fix it.


tkishel:
See also: https://gist.github.com/tkishel/710e0c4826edda63effa07c2ffba4dbd

rcaballeromx (changes requested):
Please don't move the conceptual content back into the procedure. If you feel like customers benefit from the conceptual information at specific points in the procedure, please use links to the conceptual topic.

rcaballeromx (approved):
Thanks for helping make this content more clear. PTAL at my changes.

tkishel:
LGTM!

PR #556, opened 2022-09-23
Correcting the Trusted Executable Feed description

omri86:
The Trusted Executable Feed currently states that it applies to all runtime capabilities. This is incorrect as it only applies to the file system runtime protection. Fixed the description and added the correct way of excluding binaries for other capabilities.

### DIFF
--- a/compute/admin_guide/configure/custom_feeds.adoc
+++ b/compute/admin_guide/configure/custom_feeds.adoc
@@ -78,35 +78,38 @@ It is also used immediately by the runtime defense file system sensor, which ass
 [.task]
 === Create a list of trusted executables
 
-The Trusted Executable Feed is a mechanism to address potential misidentification of legitimate files as malware. You can add to the feed signatures of binaries that have been incorrectly categorized as malicious by the file system based malware runtime protection. 
+The trusted executable list is a mechanism to address potential misidentification of legitimate files as malware. When you add signatures of binaries to this list, it enables you to ensure that a legitimate binary or a process created from a legitimate binary is not potentially identified as malicious by runtime defense rules that check for file system based malware. 
 
-NOTE: The Trusted Executable Feed does not apply to other runtime detection capabilities (e.g., process runtime protection). To exclude files from other runtime detection capabilities use the rule allowed list. Check the relevant xref:../runtime_defense/runtime_defense.adoc[runtime defense section] for more details. 
-
-You can specify one entry at a time, or do a bulk upload from a CSV file.
-The maximum file size is 20MB.
+The trusted executable list does not apply to other runtime defense capabilities, such as process runtime protection. To exclude files from other runtime detection capabilities use the allowed list in the xref:../runtime_defense/runtime_defense.adoc[runtime defense section]. 
 
 NOTE: Malware scanning and detection is supported for Linux container images and hosts only.
 Windows containers and hosts are not supported.
-
-The first line in your CSV file must be a header record that contains the field names.
-For trusted executables data, specify the MD5, followed by text description that allows you to identify why the MD5 was marked as trusted in the future. 
-Specify one entry per line.
-For example:
-
-  md5,name
-  194836fbe0f121a25b145e55e80cef22,legitimate binary built in house
-  0aeb0cac186a81a6ac45776d6b56dd70,test file
-  33cc273ae3aa8bce6a22c92e7d11f63a,benign file
 
 [.procedure]
 . Open Console.
 
 . Go to *Manage > System > Custom Feeds*.
 
-. Click *Upload Malware Data*, and either click *Add* or *Import CSV*.
+. Choose how to add the MD5 values for the trusted executables.
 +
-Your custom malware data is used in all subsequent image scans.
+The MD5 hashes you add are used in all subsequent image scans.
 It is also used immediately by the runtime defense file system sensor, which assesses all writes to the host and container file system.
+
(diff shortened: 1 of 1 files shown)


compute/admin_guide/configure/custom_feeds.adoc
```diff
@@ -78,7 +78,9 @@ It is also used immediately by the runtime defense file system sensor, which ass
 [.task]
 === Create a list of trusted executables
 
-If a legitimate binary or a process created from a legitimate binary is incorrectly identified as malicious by any of the runtime detection capabilities, you can add the signature of the binary to a list of trusted executables signatures and ensure that it will not be inspected by any runtime capabilities. 
+The Trusted Executable Feed is a mechanism to address potential misidentification of legitimate files as malware. You can add to the feed signatures of binaries that have been incorrectly categorized as malicious by the file system based malware runtime protection.
```
Pubs-MV: Hey @omri86 a couple questions:

1. is there a feed by this name "Trusted Executable Feed"? 
or are you saying that "you can create a custom feed that includes a list of trusted executables to"

2. Can we revise it to say 

If you have a list of trusted executables, you can add the signature of the binaries to a custom feed and create a trustable executables feed for your organization. This feed enables you to ensure that  a legitimate binary or a process created from a legitimate binary is not potentially identified as malicious by the file system based malware runtime protection.

omri86: 1.Yes, navigate to `system -> custom feeds -> trusted executables`
2. I don't think the revised sentence explains much better than what I wrote. It only adds confusion, because starting off with "If you have a list of trusted executables" does not explain to the customer what are those "trusted executables".

omri86: @Pubs-MV just making sure you saw my comment

Pubs-MV: Can you please review @omri86

compute/admin_guide/configure/custom_feeds.adoc
```diff
@@ -78,36 +78,39 @@ It is also used immediately by the runtime defense file system sensor, which ass
 [.task]
 === Create a list of trusted executables
 
-The Trusted Executable Feed is a mechanism to address potential misidentification of legitimate files as malware. You can add to the feed signatures of binaries that have been incorrectly categorized as malicious by the file system based malware runtime protection. 
+The trusted executable list is a mechanism to address potential misidentification of legitimate files as malware. When you add signatures of binaries to this list, it enables you to ensure that a legitimate binary or a process created from a legitimate binary is not potentially identified as malicious by runtime defense rules that check for file system based malware.
```
omri86: 1. Remove this part, it is incorrect: `or a process created from a legitimate binary`
2. Suggest to change this:
```
identified as malicious by runtime defense rules that check for file system based malware.
```
to this:
```
identified as malicious by the runtime rules file system based defense capability.
```

compute/admin_guide/configure/custom_feeds.adoc
```diff
@@ -78,36 +78,39 @@ It is also used immediately by the runtime defense file system sensor, which ass
...
 [.procedure]
 . Open Console.
 
 . Go to *Manage > System > Custom Feeds*.
 
-. Click *Upload Malware Data*, and either click *Add* or *Import CSV*.
+. Choose how to add the MD5 values for the trusted executables.
 +
-Your custom malware data is used in all subsequent image scans.
+The MD5 hashes you add are used in all subsequent image scans.
```
omri86: this change is incorrect, please leave as is or change to:
```
The MD5 hashes you add to custom malware feeds are used in all subsequent image scans.
```

compute/admin_guide/configure/custom_feeds.adoc
```diff
@@ -78,36 +78,39 @@ It is also used immediately by the runtime defense file system sensor, which ass
...
-  0aeb0cac186a81a6ac45776d6b56dd70,test file
-  33cc273ae3aa8bce6a22c92e7d11f63a,benign file
-
 [.procedure]
 . Open Console.
 
 . Go to *Manage > System > Custom Feeds*.
 
-. Click *Upload Malware Data*, and either click *Add* or *Import CSV*.
+. Choose how to add the MD5 values for the trusted executables.
```
omri86: This is inaccurate, I believe that these instructions are intended for all feeds, not just the trusted executables one.
You are also using different terms in the same doc for the same thing - MD5 hashes, MD5 values etc.

Suggest to just change the names of the buttons/links and leave text as is.

omri86 (commented):
please see my comments

compute/admin_guide/configure/custom_feeds.adoc
Pubs-MV: This section is for *Manage > System > Custom Feeds > Trusted executables".
I had missed adding the last part and have added it now.

I will use MD5 hash for consistency.

compute/admin_guide/configure/custom_feeds.adoc
Pubs-MV: This section is for trusted executables  *Manage > System > Custom Feeds > Trusted executables*

Is the reference to custom malware feed accurate here?

compute/admin_guide/configure/custom_feeds.adoc
Pubs-MV: done. broke it up to avoid  a long string of nouns.

Pubs-MV (approved):

PR #559, opened 2022-09-26
added PCSUP-11309

Pubs-MV:
adding this as a known issue in the RN as discussed on Slack and email with @Maya Even-Shani and @Or Zlociver


cfarquhar:
Thanks!

PR #582, opened 2022-10-13
[Compute AG] Agentless Fixes Fire Drill

Pubs-MV:
Fixed hard coded URls, incorporated feedback from justin and hari, all the flurry of emails

### DIFF
--- a/compute/admin_guide/authentication/credentials_store.adoc
+++ b/compute/admin_guide/authentication/credentials_store.adoc
@@ -9,7 +9,7 @@ Credentials are used when setting up the following integrations:
 * Deploying and managing Defender DaemonSets from the Console UI.
 * Injecting secrets from secret stores into containers at runtime.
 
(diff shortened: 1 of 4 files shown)


Pubs-MV (commented):
typos

harisrini:
@Pubs-MV 
Screen Shot 2022-10-13 at 12 23 16 PM
1. Move the location
2. Since this is generic, remove JSON and leave it as 
Download the permission template from...

harisrini:
@Aqsaxvv092 Post the link to the JSON and let @Pubs-MV add it directly to the note section in the above comment and also push it to the relevant Compute edition section too.

Aqsaxvv092:
@harisrini I dont have the direct JSONs. The only ones we have are for Azure from Sean S. Can we use the ones you have for SaaS M&P but just for agentless section?

PR #590, opened 2022-10-14
[Compute AG] [PCSUP-11580] Update available WildFire regions

rcaballeromx:


Pubs-MV (approved):
thanks for addressing the PCSUP. Can you please add the PCSUP number to the PR title?

PR #598, opened 2022-10-19
[RQl Ref] Replaced the tables with links to Gsheet

Pubs-MV:


Pubs-MV:
Can you please confirm that this is correct?

iansk (approved):

PR #599, opened 2022-10-19
Feature/azure service principals

jenjoe22:
Please note gsheet inegration in IAM Attributes topic.


jenjoe22:
Gsheet integration not working as expected. I will send out a new PR for the https://redlock.atlassian.net/browse/RLP-54713

PR #611, opened 2022-10-21
[CSPM AG] Fixed cross-reference links

Aishbn:
There was an issue with cross refernce links on the prod site for the following topics. Need to re-publish and chek if they work.


Pubs-MV (approved):

cspm/admin-guide/configure-external-integrations-on-prisma-cloud/integrate-prisma-cloud-with-jira.adoc
```diff
@@ -135,14 +135,22 @@ The total length of the template name can be up to 99 characters and should not
 ====
 Select the project where you want to see the Prisma Cloud alerts. Because every alert translates to a Jira ticket, as a best practice, create and use a dedicated project for Prisma Cloud ticketing and issue management.
 ====
++
+[CAUTION]
```
Pubs-MV, line 139: I think we want to make it a Note and not a Caution. Caution is for bodily harm

PR #618, opened 2022-10-26
Public sector stig update

pfox1969:
Prisma Cloud DISA STIG has been published. Cleaned up existing DISA STIG content


Pubs-MV (approved):
Thank you @pfox

PR #636, opened 2022-11-07
Update rql-examples.adoc

croy99:
added listing network ip address of the running instances ( GCP).


Pubs-MV:
Thank you @croy99 , we added your example in as part of https://github.com/PaloAltoNetworks/prisma-cloud-docs/pull/684. We addressed a whole lot of formatting fixes on this topic.

PR #638, opened 2022-11-08
Update telemetry default

grevach:
The default in the telemetry article was wrong, changed it from disabled to enabled


Pubs-MV (approved):
Thank you @grevach

PR #647, opened 2022-11-14
fix adoc syntax issues on RQL Example Library page

jjchavanne:


Pubs-MV:
was this was fixed along with your other changes @jenjoe22

jenjoe22:
> was this was fixed along with your other changes @jenjoe22

i believe we fixed this in 11.1 and merged to master already ... my edits were triggered by a query from Jim Pecora. Since I pulled from master the changes should be in there.

Pubs-MV:
Thanks @jjchavanne for being engaged.!

PR #654, opened 2022-11-16
fix-anchors

kdpanw:
hey @rcaballeromx , the hyperlinks that redirect IN page here dont work


Pubs-MV:
fixed in a separate PR.

PR #663, opened 2022-11-25
Update vuln_explorer.adoc

atkkyamasaki:
https://redlock.atlassian.net/browse/PCSUP-12655
I believe we should explain clearly that CVE result is supported only by the system admin role.


Pubs-MV (approved):
thanks @atkkyamasaki

PR #665, opened 2022-11-28
Update saved-views.adoc

marcacyr:
Add documentation about the expectation of behavior when making changes inside a Default (System) view.


Pubs-MV (approved):
Thank you @marcacyr

PR #686, opened 2022-12-02
[CWP-36091 and CWP-41843]: Updated supported.cfg and added information for the missing VM Image Scan

ssugandh:
Added and updated VM image scan files based on the issue:
https://redlock.atlassian.net/browse/CWP-36091

@iansk : I have not added the VM image scan progress endpoint. Can you please confirm if I need to add the 'progress' endpoint for VM Image scan?

We had earlier decided to remove all the progress endpoints here: https://github.com/PaloAltoNetworks/prisma-cloud-docs/pull/339 (based on https://redlock.atlassian.net/browse/PCSUP-7114)

### DIFF
--- a/api/_topic_map.yml
+++ b/api/_topic_map.yml
@@ -1383,6 +1383,10 @@
   displayName: VM image scan reports 
   description: !include ../descriptions/vms/vms.md
   
+  get:
+    summary: Get VM Image Scan Results
+    description: !include ../descriptions/vms/get.md
+
   /labels:
     get:
       summary: Get VM Image Tags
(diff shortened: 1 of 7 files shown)


Pubs-MV:
@ssugandh , based on the PCSUP-7114 let's leave out the progress endpoint as it was deprecated. @iansk , WDYT?
@ssugandh Is the endpoint still in the spec file that you download from the UI? If yes, is it marked as deprecated or is it not there any longer because it is sunset/removed to prevent use.

Pubs-MV (approved):
please see my comment earlier

PR #708, opened 2022-12-06
Removed note for requiring admin privileges to scan the registry

grevach:
Removed note for requiring admin privileges to scan the registry - After QA checked this, we do not need an admin user in order to scan the registry.
https://redlock.atlassian.net/browse/CWP-42757


sgordon46:
I would suggest you add which permissions are needed for scanning this registry.

Pubs-MV:
@grevach is a minimum privilege required?

grevach:
@Pubs-MV we didn't test all the options, but a user with read permission for the repository should be able to scan, so I'm not sure we should mention this specifically

PR #714, opened 2022-12-07
[Compute AG][GH#33333] Effects per protection in container runtime po…

divyabhushan-okta:
[GH#33333]


compute/admin_guide/runtime_defense/runtime_defense_containers.adoc
```diff
@@ -291,11 +292,13 @@ Prisma Cloud raises alerts for each of the following cases:
 * A modified binary listens on a port,
 * A modified binary makes an outbound connection.
 
+==== Learned modles
```
grevach:
```suggestion
==== Learned models
```
@divyabhushan

grevach: @divyabhushan however, I'm not sure this is the right title to use here, because the settings below are not only related to the learned models. Let's use something that is mentioning "Allowed" because this is the common part of the settings in this section

compute/admin_guide/runtime_defense/runtime_defense_containers.adoc
```diff
@@ -240,7 +241,7 @@ Also, consider using Prisma Cloud’s alert profile feature to alert specific te
 
 ==== Anti-malware
 
-Anti-malware provides high level control for anti-malware capabilities for containers. More granular configuration for each runtime capability is available through each the other tabs on the rule.
+Anti-malware provides high level control for anti-malware capabilities for containers. More granular configuration for each runtime capability is available through each of the other tabs on the rule.
 
 - *Prisma Cloud advanced threat protection* -- Use Prisma Cloud advanced threat protection intelligence feed, to apply malware prevention techniques across processes, networking and filesystem.
```
grevach, line 247: @divyabhushan we need to add a note for k8s attacks, that Prevent is not supported for Kubernetes attacks that arise from the Network flow (because Prevent is not supported for Network at all)

compute/admin_guide/runtime_defense/runtime_defense_containers.adoc
divyabhushan-okta: @grevach Thanks. How about "Allowed activities" or "Allowed processes"

grevach: @divyabhushan we can go with "Allowed activities"

Pubs-MV (approved):

PR #716, opened 2022-12-07
Correct Base Images Path

yboravelli:
In the 'Exceptions for Base Image Vulnerabilities' section, the path should be should be Defend > Vulnerabilities > Images > Base images instead of Monitor > Vulnerabilities > Images > Base images.


Pubs-MV (approved):
Thank you @yboravelli

PR #738, opened 2022-12-14
[CSPM AG] AWS topics updated post-22.12.1

arane26:
AWS topics updated for CDA based on feedback received post-22.12.1


Pubs-MV:
please confirm that the content builds and add the tested label @arane26

PR #742, opened 2022-12-14
Removed reference to data exfil case in DNS logs onboarding page

gmhoward:
The data exfil threat case is not covered with DNS logs.


Pubs-MV (approved):

Pubs-MV:
Thank you @gmhoward

PR #745, opened 2022-12-16
[Compute AG - WAAS]: Update sensitive data pattern-based rule

divyabhushan-okta:


compute/admin_guide/waas/log_scrubbing.adoc
```diff
@@ -21,7 +21,7 @@ image::./waas-sensitive-data.png[width=350]
 
 . For pattern-based rules:
 +
-image::./waas_log_scrubbing_new_rule_dialog.png[width=300]
+image::./waas_log_scrubbing_new_rule_dialog.png[15%]
```
Pubs-MV: does using `% ` work?

I thought we need the `scale` instead of `width`

and the value as an integer

divyabhushan-okta: Yes, both work. Yes, I changed it to `scale` as in other places. Thank you.

Pubs-MV: Did % work on stage? I do not believe it will work, please use scale =. 
Ian mentioned that we have a hard coded value in our backend. So, the % your are adding is not being honored. 

@iansk please confirm.

divyabhushan-okta: Sure. I changed it to `scale` already in the previous commit.

iansk: Currently, specifying `width` or `scale` doesn't have any effect. The converter hard-codes a scale of 70%. I'll fix this in the converter shortly, and then `scale` will work as expected.

PR #748, opened 2022-12-19
Updated webhook alert macros

grevach:
The macros in the article weren't updated, added new macros missing


Pubs-MV (approved):
Thank you @grevach

PR #755, opened 2022-12-21
[Compute AG] Fix broken link from WAAS

Pubs-MV:
Fixing to publish OCI changes for RC.


Pubs-MV:
we need a way to catch these @iansk

iansk:
@Pubs-MV Agreed. Created PCDTUS-30 to track it

PR #777, opened 2023-01-06
[CSPM AG] Separate static CFT links 

Pubs-MV:


Pubs-MV:
@sridharam

Pubs-MV:
@arane26, we looked at this together.

PR #780, opened 2023-01-09
[Compute AG: CWP-42222]: Update VPC troubleshooting

divyabhushan-okta:
Reference: https://redlock.atlassian.net/browse/CWP-42222


compute/admin_guide/book_compute_edition.yml
```diff
@@ -541,7 +541,7 @@ topics:
     file: deployment_serverless.adoc
   - name: Deploy WAAS Out of Band
     file: deployment_out_of_band.adoc
-  - name: Deploy WAAS Out of band with VPC Traffic Mirroring
+  - name: Deploy WAAS Agentless
```
Pubs-MV, line 544: @divyabhushan , is WAAS agentless the same as OOB or only OOB with VPC traffic mirroring?

divyabhushan-okta: Hi @Pubs-MV, WAAS Agentless is same as "OOB with VPC Traffic mirroring" without a Defender (The user doesn't have to install a Defender, VPC configuration will auto-install an Observer on EC2 to monitor the traffic OOB from the source).
The other type 'WAAS OOB' monitors traffic OOB and the user has to install a Defender.

Pubs-MV (approved):

PR #789, opened 2023-01-11
[Compute AG] Remove links to the Twistlock CDN

iansk:
Fix links to the Compute API. Links to downloads/attachments remain as-is.

### DIFF
--- a/compute/admin_guide/audit/event_viewer.adoc
+++ b/compute/admin_guide/audit/event_viewer.adoc
@@ -1,7 +1,7 @@
 == Event viewer
 
 Prisma Cloud creates and stores audit event records (audits) for all major subsystems.
-Audits can be reviewed in *Monitor > Events*, or they can be retrieved from the https://cdn.twistlock.com/docs/api/twistlock_api.html[Prisma Cloud API].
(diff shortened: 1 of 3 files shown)


compute/admin_guide/audit/event_viewer.adoc
```diff
@@ -1,7 +1,7 @@
 == Event viewer
 
 Prisma Cloud creates and stores audit event records (audits) for all major subsystems.
-Audits can be reviewed in *Monitor > Events*, or they can be retrieved from the https://cdn.twistlock.com/docs/api/twistlock_api.html[Prisma Cloud API].
+Audits can be reviewed in *Monitor > Events*, or they can be retrieved from the https://prisma.pan.dev/api/cloud/cwpp/[Prisma Cloud Compute API].
```
Pubs-MV: can we link to https://prisma.pan.dev/api/cloud/cwpp/audits instead?

iansk: @Pubs-MV Good idea. Fixed.

iansk:
Merging this now. This PR fixes links to the API docs, which needs to be fixed immediately because the API docs on the CDN are outdated and they've been taken down.

Fixes to attachments on the CDN to be fixed in a separate PR later.

PR #798, opened 2023-01-13
Update prisma-cloud-licenses.adoc

atkkyamasaki:
Kindly refer to the following JIRA.
https://redlock.atlassian.net/browse/PCSUP-13618

=== HERE ===
Would it be possible to ingest public cloud resources (The MDC work or stop) ?
Ingestion does not run when license is expired.


Pubs-MV:
Yes, we can document the behavior on license expiry.

Pubs-MV:
content was updated as part of a different PR.

PR #821, opened 2023-01-19
[Compute AG] Update System Requirements for Lagrange Update 1

rcaballeromx:


grevach:
Hi @rcaballeromx 
I am not sure if setting Talos Linux support under "supported orchestrators" is the right location. This is an OS so maybe we should place it under supported OSs

Pubs-MV:
@rcaballeromx I am moving it under the Supported Operating Systems on x86_64.

PR #826, opened 2023-01-22
[Compute AG] Moved Talos Linux to Host OS section

Pubs-MV:
and fixed note

### DIFF
--- a/compute/admin_guide/install/system_requirements.adoc
+++ b/compute/admin_guide/install/system_requirements.adoc
@@ -201,8 +201,11 @@ The following features are not available for Talos OS:
 * Scanning of underlying hosts
 * Runtime scanning
 * Agentless scanning
-* Automatic recognition of the cluster name
-* Block policies
+* Vulnerability and compliance blocking policies
+* Automatic recognition of the cluster name.
++
+You must enter it manually in the *Cluster name* field when deploying Defenders or twistcli
+
 ====
 
 |Ubuntu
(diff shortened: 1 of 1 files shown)


compute/admin_guide/install/system_requirements.adoc
```diff
@@ -185,6 +185,21 @@ Runtime prevent capability is supported only for DNS events. Other prevent capab
...
 SLES 15  SP1 - SP4 - Only host Defenders are supported.
 
+|Talos OS
+|Talos OS 1.3.0
+Kernel version: 5.15.83-talos
+Kubernetes version: 1.25.4
+containerd version: 1.6.12
+
+
+The following features are not available for Talos OS.
```
grevach: @Pubs-MV do we want to set this as a note? similar to bottlerocket

Pubs-MV: done, and also clarifying that we do not support both Vuln and Compliance block policies

compute/admin_guide/install/system_requirements.adoc
```diff
@@ -185,6 +188,26 @@ Runtime prevent capability is supported only for DNS events. Other prevent capab
...
+====
+The following features are not available for Talos OS:
+
+* Scanning of underlying hosts
+* Runtime scanning
+* Agentless scanning
+* Vulnerability and compliance blocking policies
+* Automatic recognition of the cluster name.
++
+You must enter it manually in the *Cluster name* field when deploying Defenders or twistcli
```
grevach:
```suggestion
You can enter it manually in the *Cluster name* field when deploying Defenders via UI or twistcli
```

Pubs-MV: done

PR #832, opened 2023-01-23
Fixed typo

arnaudpion:
setting hte effect > setting the effect


Pubs-MV (approved):
Thank you @arnaudpion

PR #835, opened 2023-01-24
Add minimum python version for checkov

mikeurbanski1:


Pubs-MV (commented):
@mikeurbanski1 Thanks for the PR. Does this change work for you?

I added the Python requirement in step 2

PR #843, opened 2023-01-26
[Compute AG] Remove specific maintenance release for GKE

rcaballeromx:


Pubs-MV:
Fix for the issue from email for GKE version ( patches do not matter)

PR #856, opened 2023-01-30
Update agentless-scanning.adoc

omri86:
1. Let's avoid mentioning specific limited capabilities on the front page and leave it to a per CSP guidance.
2. Let's not list random limitations on the opening landing page of the feature, let's leave these limitations for SAs to handle, we have many more limitations that we don't list here, so I don't find it helpful to list just 3 random ones.
2. Hub account is only for AWS and GCP, in current structure it implies that we only support AWS and GCP generally - let's drop this altogether


rcaballeromx (approved):
I'm good with the proposed changes now that we have added the CSP-specific topics.

omri86:
@rcaballeromx we need this changes merged to the PCEE edition as well

PR #857, opened 2023-01-31
Add tag description and example

ruzickap:
Add better description, how to use tags in resource lists.


Pubs-MV (approved):

Pubs-MV:
Thank you @ruzickap

PR #873, opened 2023-02-03
[Compute AG: PCSUP-13847]: Add CRI-O images scan exception

divyabhushan-okta:
Reference: https://redlock.atlassian.net/browse/PCSUP-13847


Pubs-MV:
vuln management topic removed from PCCE and PCEE

PR #874, opened 2023-02-03
Serverless functions for agentless - added

georgmarkarian:
To have complete visibility of agentless support across hosts, container and serverless. serverless function is added to the doc


compute/admin_guide/agentless-scanning/agentless-scanning.adoc
```diff
@@ -3,7 +3,7 @@
 Agentless scanning lets you inspect the risks and vulnerabilities of a cloud workloads without having to install an agent or affecting the execution of your workload.
 Prisma Cloud gives you the flexibility to choose between agentless and agent-based security using Defenders.
 Prisma Cloud supports agentless scanning cloud workloads on AWS, Azure, GCP, and OCI for vulnerabilities and compliance.
-In AWS, Azure, and GCP you can also use agentless scanning for container images.
+In AWS, Azure, and GCP you can also use agentless scanning for container images as well as xref:../vulnerability_management/serverless_functions.adoc[Serverless functions].
```
rcaballeromx, line 7: @georgmarkarian Do we support serverless scanning with agentless for GCP? This change implies that and I'm not sure we do.

rcaballeromx (changes requested):

compute/admin_guide/agentless-scanning/agentless-scanning.adoc
georgmarkarian: we support serverless scanning with AWS, Azure and GCP, which is also agentless, the difference is that the scanners are not deployed in customer cloud account for serverless scanning but the prisma console is used for that. This is to confirm that agents are not required to scann serverless functions

rcaballeromx (approved):
I've resolved the merge conflict and will merge the fix. Thanks

PR #879, opened 2023-02-06
[Compute AG] Fix required GCR Scan IAM role -CWP-45521

Pubs-MV:
The required IAM role is called `Storage Object Viewer`.


Pubs-MV:
CWP-45521

PR #882, opened 2023-02-07
clarify user_roles.adoc

kdpanw:
this shows up in the saas docs, but none of these roles are relevant to SaaS in any way. None of them are available. We need to clarify that piece and/or remove this page entirely as it has no value technically for SaaS deployments. Perhaps PM team or platform team can comment or something


Pubs-MV:
there is a PR in progress: https://github.com/PaloAltoNetworks/prisma-cloud-docs/pull/878 for https://redlock.atlassian.net/browse/CWP-45450

PR #883, opened 2023-02-07
Update integrate-prisma-cloud-with-aws-id-center.adoc

ipalvr:
The RQL query should be - config from iam where source.idp.service = 'AWS Identity Center'


Pubs-MV (commented):
Thanks @ipalvr

PR #931, opened 2023-02-17
Update drift-detection.adoc

akinpanw:
Customers are not aware of this automated drift detection for Cloudformation. Just added this for clarity. If in the right location, we can move it around. Thank you.


Pubs-MV (commented):
Thanks @akinpanw

PR #934, opened 2023-02-20
Add more iam query examples

jjchavanne:
- Added more IAM query examples
- Added sections in the Azure IAM query section to make easier to read and find relevant examples
- Fixed broken internal page ref link to General IAM Examples section


Pubs-MV:
Thanks @jjchavanne !

PR #936, opened 2023-02-20
[Compute AG - CWP-40557]: Add note for ACR credentials

divyabhushan-okta:
Reference: https://redlock.atlassian.net/browse/CWP-40557

Screenshot 2023-02-20 at 10 17 42 PM


rcaballeromx (approved):
Great work.
@Pubs-MV we can publish this change now and revisit when the credentials are imported automatically from CSPM.

PR #939, opened 2023-02-21
[Compute AG]: Update Prisma.pan.dev references to Pan.dev

ssugandh:
Urls adjusted for the new pan.dev API reference site.

To be merged when the pan.dev turns live.


Pubs-MV:
needs to be merged with  #941

PR #941, opened 2023-02-22
[CSPM AG] replaced path from prisma.pan.dev

Pubs-MV:


Pubs-MV:
needs to be merged with #939

PR #951, opened 2023-02-24
[CSPM RN] - Removed API ingestions from LA 

Aishbn:
### Removed API ingestions from LA that will be going with 23.2.2

The following API ingestions were removed based on a late request from the QA team. 

- AWS Network Firewall - RLP-92179
- AWS Systems Manager - RLP-92175


arane26 (approved):
ok to merge

PR #963, opened 2023-03-01
Update access-prisma-cloud.adoc

priteshchandaliya:
The legacy login URL was incorrect


Pubs-MV:
Thanks @priteshchandaliya

PR #976, opened 2023-03-07
[CodeSec AG] Update line to indicate single support for TF [RLP-95052]

anaghapamidi:
Summary of Changes:

From Dganit. There is a regression for Terraform Cloud Run Task and Terraform Enterprise Run Task that is currently being worked on, therefore rephrasing the sentence that states "a single account that support multiple integrations."

Modified Files:

- add-terraform-run-tasks.adoc
- add-terraform-enterprise-run-tasks.adoc
- add-terraform-enterprise.adoc
- tf-enterprise-13.png
- tf-enterprise-10.png


Pubs-MV:
why did we chose to update the AG instead of adding this as a known issue in the Release Notes @anaghapamidi @tsmithv11

PR #991, opened 2023-03-09
[Compute RN] Remove incorrect Backwards Compatibility entry

rcaballeromx:


Pubs-MV:
Removing 22.01 entry after confirming with AH over email.

PR #997, opened 2023-03-10
Update install_serverless_defender.adoc

jjeanclaude:
According to the document below, NodeJS 16 and 18 are supported

https://docs.paloaltonetworks.com/prisma/prisma-cloud/prisma-cloud-admin-compute/install/system_requirements (Serverless Runtimes)

If that's the case, those should be specified in this document


rcaballeromx:
Closing this PR because I've included the fix in the Defender restructure work. #925 
The fix is being published with the content for Maxwell.

PR #1029, opened 2023-03-23
[Compute API] v1-corrections-cURL-examples

ssugandh:
There are total 35 instances of v1 in supported endpoints cURL examples that needed to be changed to vVERSION.


Pubs-MV:
for  https://redlock.atlassian.net/browse/CWP-47032

Pubs-MV (approved):

PR #1061, opened 2023-04-04
replacing with the new screenshot. 

priteshchandaliya:
The old screenshot says that we only support AWS, but now we support IAM for all the CSPs


Pubs-MV (commented):
thank you @priteshchandaliya

PR #1081, opened 2023-04-11
update vm_images.adoc

kdpanw:
Add a note/blurb that current behavior will halt an entire scan if 1 image in the scope fails to scan .This is current behavior , we get cases on it and it's a feature level fix but it's not made aware to customers in any way.

I don't know the best place to add this content here-


kdpanw:
@Pubs-MV I didn't make a change as I don't know the best place for this information on the page, just added a whitespace to allow the PR to be made. Perhaps your team can decide best place for this info

rcaballeromx:
@kdpanw We will look into it and will draft something. I'm adding labels to the PR to avoid problems during release.

PR #1090, opened 2023-04-12
Update prisma_cloud_vulnerability_feed.adoc

mklambert:
The contents of the IS feed are proprietary and should not be referenced directly. I've removed the reference to that data.


Pubs-MV (approved):
Thank you @mklambert

PR #1111, opened 2023-04-20
[CSPM] AG AWS and GCP simplified onboarding updates

arane26:
AWS, GCP, and PCDS topics and images updated for simplified onboarding in 23.4.2
Old AWS and GCP topics removed from book.yml


Pubs-MV:
Confirmed that we do not merge this PR with @arane26. She will submit a separate PR

PR #1126, opened 2023-04-24
[RQL Reference] Image-cleanup

jenjoe22:


Pubs-MV (approved):
yes.. the PDF builds without any errors @jenjoe22

PR #1134, opened 2023-04-25
[Compute AG] link to linux man page

Pubs-MV:


Pubs-MV:
@iansk , how do we need to format the section on Kermel capabililties to prevent the "link" to the same topic.

See HTML for what I mean
https://docs.paloaltonetworks.com/prisma/prisma-cloud/30/prisma-cloud-compute-edition-admin/install/system_requirements#kernel

iansk:
@Pubs-MV I don't see/understand the issue. Please provide more details

iansk:
Context from Slack:

The issue is that  "[CAP_NET_ADMIN](https://docs.paloaltonetworks.com/prisma/prisma-cloud/30/prisma-cloud-compute-edition-admin/install/system_requirements#)" on the Kernel Capabilities section of this [page](https://docs.paloaltonetworks.com/prisma/prisma-cloud/30/prisma-cloud-compute-edition-admin/install/system_requirements), displays as a link.

The link is to the same page.

sys-reqs-kernel-caps-links

None of the highlighted items should be links. I have added a link to the Linux man page in the lead-in sentence

iansk:
@Pubs-MV I took a closer look. Our tools are outputting valid DITA XML:

```
<ul>
   <li><p><term>CAP_NET_ADMIN</term></p></li>
   <li><p><term>CAP_NET_RAW</term></p></li>
   ...
</ul>
```

This looks like a bug in how AMS handles the DITA XML to HTML transform for `term` (and may be unique to when `term` is inside a list item inside a `ul`). For some reason, it's being transformed into a link in the HTML. In the PDF, it's transformed correctly. In the PDF, it's just italics text.

HTML:

dita-xml-transform-term-to-link

PDF:

dita-xml-transform-term-to-link-pdf

If you want to get rid of the links, I suggest getting replacing the `term` markers (backticks) and make the relevant text bold instead. It's a decent work-around. And given that we might be moving off this system, not worth spending more time on it.

Pubs-MV:
Thanks for the investigation @iansk

PR #1155, opened 2023-05-02
Ag-image-clean-up

jenjoe22:
deleted unused images from cspm-ag .... batcg tested by successfully building with each commit.


Pubs-MV:
Thank you for this work @jenjoe22 !

Pubs-MV:
PDF builds without any errors wrt to these images

Pubs-MV (approved):

PR #1167, opened 2023-05-05
[Compute AG: PCSUP-11710]: Update redeploy defender command

divyabhushan-okta:
## Reference

[PCSUP-11710](https://redlock.atlassian.net/browse/PCSUP-11710)

### Redeploy Defenders
Screenshot 2023-05-05 at 9 53 14 AM

### Create a new Custom Check
Screenshot 2023-05-16 at 12 23 41 PM

### DIFF
--- a/compute/admin_guide/compliance/custom_compliance_checks.adoc
+++ b/compute/admin_guide/compliance/custom_compliance_checks.adoc
@@ -34,6 +34,7 @@ As new custom checks are added, they are automatically assigned the next availab
 [NOTE]
 ====
 Prisma Cloud drops the cached compliance and vulnerability scan results for registries, and rescans registry images, whenever:
+
 * A new rule referencing a custom compliance check is added, or
 * An existing compliance check referenced by some existing rule is updated, or
 * An existing rule is updated with a new custom compliance check.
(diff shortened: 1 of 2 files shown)


compute/admin_guide/compliance/custom_compliance_checks.adoc
```diff
@@ -56,7 +56,7 @@ You only need to redeploy Defenders when switching to the enabled state.
 . Set *Custom Compliance Checks for hosts* to enabled.
 
 . Deploy Defenders to your environment.
-Or if already deployed, redeploy your Defenders.
+Or if already deployed, xref:../install/deploy-defender/redeploy-defender.adoc[redeploy your Defenders].
```
Pubs-MV: @divyabhushan this link here is good.

Can you please confirm with QA ( Carlos or Alex) if we also need to redeploy Defenders when you create a new custom check (workflow below) and add it to the compliance policy?

compute/admin_guide/install/deploy-defender/redeploy-defender.adoc
```diff
@@ -20,9 +20,16 @@ $ <PLATFORM>/twistcli defender export kubernetes \
     --cluster-address twistlock-console
 ----
 
-. Apply the https://kubernetes.io/docs/concepts/cluster-administration/manage-deployment/#in-place-updates-of-resources[in-place updates] to your `Defender` resources.
+. Delete the old Defenders using your old daemonset config file:
```
Pubs-MV, line 25: Please confirm that with QA also. The main reason is that does deleting the Defenders contradict the next step where we talk about "in-place updates".

compute/admin_guide/compliance/custom_compliance_checks.adoc
divyabhushan-okta: Hi @Pubs-MV, 
As per the comments from Tal over [here](https://redlock.atlassian.net/browse/PCSUP-11710?focusedCommentId=538969), the workflow is that the custom check has to be enabled first, the defenders redeployed next, and then create the custom check.
I believe, we should rewrite this page to include the steps to enable the setting and redeploy defenders as a prerequisite step before creating a new custom check.

divyabhushan-okta: @Pubs-MV
So, I rearranged the prerequisite section as we discussed over Slack and inline with Tal's comments above.

compute/admin_guide/install/deploy-defender/redeploy-defender.adoc
divyabhushan-okta: @Pubs-MV validated on the ticket.

PR #1175, opened 2023-05-10
fix curl example for CSPM authentication

beahues:
single dash `-data` was being handled as `-d "ata"`, rather than the correct `--data <argument>`.


Pubs-MV:
thank you @beahues

Pubs-MV (approved):

PR #1179, opened 2023-05-11
[Compute AG] Add Agentless Scanning concept

rcaballeromx:


compute/admin_guide/agentless-scanning/understand-agentless-scanning.adoc
```diff
@@ -0,0 +1,83 @@
+:toc: macro
+== Understanding Agentless Scanning of Workloads
```
Pubs-MV: We do not use "Understanding" or "Overview" in our style.

2 questions: 

1. Do we need a separate topic? The information in this topic seems apt for the agemtless-scanning.adoc file.

2. If you believe we need this topic, can we love the heading called "Understanding" and name it "Agentless Scanning Process" or "How Agentless Scanning Works"

I understand that we will have an "orphan" heading for Networking Infrastructure, but hopefully we will add more later.

Pubs-MV (approved):

PR #1195, opened 2023-05-16
[Compute AG - Email Alerts]: Fix image

divyabhushan-okta:
Rename the image name


Pubs-MV (commented):
thank you @divyabhushan

PR #1220, opened 2023-05-19
Update install-gke-autopilot.adoc

pasqua1e:
We need to remove port 8083, because this is not self hosted. We also need to add port 443 to the cluster address (which is used for the defenders to point to the console via WSS protocol).

Finally, we should add a note to explain where this console address is coming from


Pubs-MV:
another one for PCEE vs PCCE

Pubs-MV:
@rcaballeromx , you may have missed this as I had assigned it to you, and had not added a comment.

rcaballeromx:
I created PR #1242 to add the conditional. I'm closing this PR.

PR #1221, opened 2023-05-19
Update install_kubernetes_cri.adoc

pasqua1e:
we should remove ports 8083/8084 as they refer to the self-hosted version


Pubs-MV:
@rcaballeromx can you please reveiw this and update for both PCCE and PCEE. Would it make sense to add the port numbers as ifdef and not inline in the example?

rcaballeromx:
@pasqua1e I've added the ports for PCEE in #1241 Thank you for providing the ports. We need to publish both and I added the conditional for each version. Thus,  I'm closing this PR.

PR #1248, opened 2023-05-26
Update create-a-policy.adoc

atkkyamasaki:
The customer reported the description is not clear. So I suggest changing it.
(FYI)
https://redlock.atlassian.net/browse/PCSUP-16458


Pubs-MV (approved):

Pubs-MV:
Thank you @atkkyamasaki

PR #1272, opened 2023-06-02
Update bc-k8s-17 and bc-k8s-18

tsmithv11:
Change fix suggestions to explicit denies based on user feedback that it's unclear.


anaghapamidi:
Dita tested.

anaghapamidi (approved):

PR #1288, opened 2023-06-11
[Code Sec] Update Bitbucket integration docs for webhooks and permissions

tsmithv11:
Requested by a customer to make it clear why we need the permissions we need and that a webhook is automatically created


Pubs-MV (approved):
Thanks @tsmithv11

PR #1290, opened 2023-06-13
update to include more enforcement level information in docs for TF cloud and enterprise

epartington:
took enforcement levels from here and updated doc with more info https://bridgecrew.io/blog/tfe-integration-policy-as-code-at-scale/#:~:text=%60hard%2Dmandatory%60%20means%20your,to%20still%20apply%20the%20IaC


Pubs-MV:
Thanks @epartington @anaghapamidi

PR #1291, opened 2023-06-13
[CSPM RN] update Severity from Medium to High for policy

jluevan13:
update severity for AWS EC2 instance that is reachable from untrust internet source to ports with high risk. 

This policy has a severity of High in the console so updating the rn to match the policy severity.


Pubs-MV (approved):
Thank you @jluevan13

PR #1298, opened 2023-06-15
Update operationalize.adoc

okostine-panw:
Update CNNS and WAAS


Pubs-MV:
@okostine-panw this is guide is not published anymore. Did you take the trouble to find the source file on GH and edit it?

PR #1310, opened 2023-06-15
syntax error

jenjoe22:


Pubs-MV:
@anaghapamidi , please review this topic as there are several notes that are not in our style. Example: the Add Base URL step. 
There is no need for a note, we can make it part of the explanation to say that "The Base URL that you enter must begin with https://.

Thanks for the fix @jenjoe22

PR #1311, opened 2023-06-19
[CSPM RN] LA Fixes post 6.2 publish

Aishbn:


Pubs-MV:
was this removed by mistake in 6.2  @Aishbn ?

Aishbn:
@jenjoe22 , can you look at @Pubs-MV above comment and respond please?

jenjoe22:
Yes I moved this to the New Features in the first draft of the RN and then we got guidance that it can remain in the LA and I think I  removed it from the RN but did not move it back to the LA. Thanks for correcting it Aish.

PR #1318, opened 2023-06-22
Update v30_01_153.adoc

eric-bischoff:
Updated Justifications


eric-bischoff:
@Pubs-MV Quick update to the 30.01.153 DISA STIG Justifications. This is now finalized.

Pubs-MV (approved):

PR #1334, opened 2023-06-27
[Compute AG] Updates for CWP-47373

Pubs-MV:
and removed unnecessary sections


Pubs-MV:
Approved by Alex H over Jira.

PR #1348, opened 2023-06-29
[CodeSec AG] Update for 23.7.1

anaghapamidi:
Summary of Changes:

Modified files:

- scan-monitor/monitor-fix-issues-in-scan/monitor-fix-issues-in-scan.adoc
- proj-16.png

PR #1351, opened 2023-06-29
Update add-pre-receive-hooks.adoc

Pubs-MV:
fixes because there is lead in content that is not a task. Therefore topic type is not a task


Pubs-MV:
fixes for [PR](https://github.com/PaloAltoNetworks/prisma-cloud-docs/pull/1331) and https://github.com/PaloAltoNetworks/prisma-cloud-docs/pull/1348

@anaghapamidi

PR #1371, opened 2023-07-10
CWP-49652: Update registry_post.md for GitLab Container Registry example and details

ssugandh:
@Pubs-MV : I have added details for the new schema in the API docs also. 
Please review and merge for Maxwell Update 3.

### DIFF
--- a/api/descriptions/settings/registry_post.md
+++ b/api/descriptions/settings/registry_post.md
@@ -10,21 +10,24 @@ It can be one of the following strings:
 
 * Amazon EC2 Container Registry: `aws`
 * Azure Container Registry: `azure`
+* CoreOS Quay: `coreos`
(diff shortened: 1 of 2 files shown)


api/descriptions/settings/registry_post.md
```diff
@@ -45,3 +48,41 @@ curl -k \
...
+  -H 'Content-Type: application/json' \
+  -X POST \
+  -d '
+  {
+    "version":"gitlab",
+    "registry":"https://registry.gitlab.com",
+    "namespace":"",
+    "repository":"",
+    "tag":"",
+    "credentialID":"<GitLab Token>",
```
tallieber, line 78: This is not correct: 
It should be the credential ID  of the credential the user created under credential store. Should be fixed before release

PR #1374, opened 2023-07-11
Update define-prisma-cloud-enterprise-settings.adoc

Ivani451:
Customer is getting confused on the wording used and mentioned that the edited sentence sounds contradictory to the previous one. Using "auto-enabled" clears up the confusion, in my opinion.


Pubs-MV (approved):

Pubs-MV:
Thanks @Ivani451

PR #1379, opened 2023-07-12
Update deployment_app_embedded.adoc

pasqua1e:
I'd like to propose some changes to the WAAS app-embedded doc, given the particular nature of the policies. We need to clarify that the auto-detect option is not available. There are 2 different/conflicting screenshots here. Also, we should remove the Host (Linux/Win) comments

### DIFF
--- a/compute/admin_guide/waas/deploy_waas/fragments/add-app-policy.adoc
+++ b/compute/admin_guide/waas/deploy_waas/fragments/add-app-policy.adoc
@@ -1,6 +1,6 @@
 . Select a WAAS rule to add an App in.
 
-. Click *Add app*.
+. Select *Add app*.
 
 . In the *App Definition* tab, enter an *App ID*.
 +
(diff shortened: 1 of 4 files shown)


compute/admin_guide/waas/deploy_waas/fragments/add-app-policy.adoc
```diff
@@ -9,23 +9,24 @@ NOTE: The combination of *Rule name* and *App ID* must be unique across In-Line
...
 
 .. Enter *HTTP host* (optional, wildcards supported).
 +
 HTTP hostnames are specified in the form of [hostname]:[external port].
 +
 The external port is defined as the TCP port on the host, listening for inbound HTTP traffic. If the value of the external port is "80" for non-TLS endpoints or "443" for TLS endpoints it can be omitted. Examples: "*.example.site", "docs.example.site", "www.example.site:8080", etc.
 
-.. Enter *App ports* as the internal port your app listens on (optional, if you selected *Automatically detect ports* while creating the rule). 
+.. Enter *App ports* as the internal port your app listens on.
+ifndef::waas-app-embedded[]
```
divyabhushan-okta, line 24: This condition excludes the following lines from appearing in the App-Embedded WAAS page.

PR #1381, opened 2023-07-12
[CSPM AG] 23.7.2 updates for UAI and drift alerts

arane26:
ag updates and fixes for 23.7.2


Pubs-MV:
confirmed with Ashwini that I can remove the DNM label

PR #1388, opened 2023-07-13
Update BB docs to add value prop and screenshots

tsmithv11:
Customer asked to add the value prop to the top of these pages and to update the screenshots to reflect the proper permissions (before they were both read which is inaccurate).


Pubs-MV (approved):

Pubs-MV:
Thank you @tsmithv11

PR #1427, opened 2023-07-28
[CSPM RN} added attack path policy renaming to look ahead

Pubs-MV:


Pubs-MV:
cc @Aishbn this came in today

PR #1428, opened 2023-07-29
Fix wrong argument on CKV_AZURE_139 policy

bruno-motacardoso:
It seems that the argument specified on the guideline is wrong.
I took advantage to improve the description of the fix.


Pubs-MV:
Can you please confirm that this change is accurate @tsmithv11

tsmithv11 (approved):
Looks good

PR #1476, opened 2023-08-16
[Compute AG: CWP-42722]: certificate info for webhook

divyabhushan-okta:
## Reference

https://redlock.atlassian.net/browse/CWP-42722


compute/admin_guide/vulnerability_management/registry_scanning/webhooks.adoc
```diff
@@ -26,11 +26,13 @@ Prisma Cloud scans images in response to layer push events.
 
 For Azure Registry, you can configure webhooks for your container registry that generate events when certain actions are performed against it. See https://docs.microsoft.com/en-us/azure/container-registry/container-registry-webhook-reference/[Azure's documentation] for more information.
 
+The Console needs a valid and trusted certificate for the webhook to work on JFrog SaaS server.
```
Pubs-MV: @divyabhushan what does SaaS server mean here?

Do you mean JFrog Cloud? 

Also please  take a look at this sentence "The benefit of webhook-initiated scans is that they are triggered as soon as images change, but support is limited to Docker Hub, Docker Registry, and Azure Registry." The list seems incomplete and maybe we can revise it to remove the listing beacuse we have that above.

divyabhushan-okta: @Pubs-MV
Yes, I renamed it to JFrog Cloud. Also, removed the repeated webhook support line.

PR #1457, opened 2023-08-10
[AppSec RN] 23.8.2

anaghapamidi:

PR #1492, opened 2023-08-23
[App Sec GA/CSPM RN]

Pubs-MV:
merged the changes that @anaghpamidi had in #1457 that I had reverted in #1481.

Then moved and revised stuff that was just lying around.


anaghapamidi (approved):
Have committed the changes.

PR #1502, opened 2023-08-28
[Compute AG: CWP-51107]: Update binary to app package type

divyabhushan-okta:
## Reference

https://redlock.atlassian.net/browse/CWP-51107


tallieber:
@divyabhushan We pause this change because we didnt have enough time to test it, so currently won't merge

divyabhushan-okta:
@tallieber Thanks for letting me know! Would you please update the Jira ticket "Fix versions"!
Thanks

PR #1505, opened 2023-09-01
DISA STIG Release Findings TechDoc Structure Update 

eric-bischoff:
DISA STIG Release Findings TechDoc Structure Update


eric-bischoff:
@Pubs-MV Made some structure changes to the DISA STIG TechDocs. Can you confirm that the graphics are in line for Prisma Cloud TechDocs? (Blue header etc). Thanks!

PR #1572, opened 2023-10-01
[CSPM AG] PCSUP 12938 - Update  Scheduled descirption prisma-cloud-al…

Izabellaya:
…ert-resolution-reasons.adoc


Pubs-MV (approved):
Thank you @Izabellaya

PR #1590, opened 2023-10-06
Remove unneeded host dependencies from localhost pre-receive setup 

james-otten-pan:
Remove 2 links for things to install before the localhost install. Neither are needed because the steps run the script inside of the image, not on the host.

Also, one of the links 404s.


tsmithv11 (approved):
You're right. Must have been a leftover from when we started direct. 🍻 

@kartikp10 FYI

Pubs-MV (approved):

Pubs-MV:
thank you @james-otten-pan

PR #1592, opened 2023-10-08
Update Update the section about unsupported Azure IAM concepts in  CS…

Izabellaya:
…PM AD context-used-to-calculate-effective-permissions.adoc

Update Update the section about unsupported Azure IAM concepts in  CSPM AD context-used-to-calculate-effective-permissions.adoc


Pubs-MV (approved):

Pubs-MV:
Thank you @Izabellaya 

cc @jenjoe22

PR #1631, opened 2023-10-17
search-investigate, alerts update

arane26:
added missing images, fixed typos, updated risk prioritization and added xrefs to alerts


arane26:
ran panconv to test mini books and updated main book

